PR #3, opened 2024-12-03
fixed #2. create home in /opt/.

### LINKED ISSUE #2: Do not create a new user with a home folder at /home/fivenines/
Maybe use /opt/fivenines/ as other user-installed applications do?

```
⚠️  Note: '/home/fivenines/.local/bin' is not on your PATH environment variable. These apps will not be globally accessible until your PATH is updated.
....
$ sudo find /home/fivenines/|tail
/home/fivenines/.cache/pip/http/b/8/9
/home/fivenines/.cache/pip/http/b/8/9/2
/home/fivenines/.cache/pip/http/b/8/9/2/5
/home/fivenines/.cache/pip/http/b/8/9/2/5/b8925380a8a772284457df6a52b6ce5b5d73a9613f9572fd67ac93d7
/home/fivenines/.cache/pip/http/a
/home/fivenines/.cache/pip/http/a/1
/home/fivenines/.cache/pip/http/a/1/9
/home/fivenines/.cache/pip/http/a/1/9/5
/home/fivenines/.cache/pip/http/a/1/9/5/3
/home/fivenines/.cache/pip/http/a/1/9/5/3/a19537d3cf37c122db841d6fe4cd322bc10d1a558bb00d146b85cb9a
```

bluet:
current installer creates user home at /home/fivenines/, but it would better to be at /opt/fivenines/.

fixes #2

### DIFF
--- a/fivenines_setup.sh
+++ b/fivenines_setup.sh
@@ -29,7 +29,7 @@ echo -n "$1" | sudo tee /etc/fivenines_agent/TOKEN > /dev/null
 
 # Create a system user for the agent
 if ! id -u fivenines >/dev/null 2>&1; then
-  sudo useradd --system --user-group --key USERGROUPS_ENAB=yes fivenines --shell /bin/false --create-home
+  sudo useradd --system --user-group --key USERGROUPS_ENAB=yes fivenines --shell /bin/false --create-home -b /opt/
 fi
 
 # Determine the package manager and install dependencies
--- a/fivenines_update.sh
+++ b/fivenines_update.sh
@@ -1,4 +1,45 @@
 #!/bin/bash
+# This script is used to update the fivenines agent, and run by root
+
+# stop the agent
+sudo systemctl stop fivenines-agent.service
+
+# if the home directory of user "fivenines" is /home/fivenines (which is the old location), migrate user's home directory to /opt/fivenines
+if [ "$(getent passwd fivenines | cut -d: -f6)" == "/home/fivenines" ]; then
+        echo "Migrating fivenines.io's working directory from /home/fivenines to /opt/fivenines"
+        # if /opt/fivenines exists, or /home/fivenines not exists, exit
+        if [ -d /opt/fivenines ] || [ ! -d /home/fivenines ]; then
+                echo "Error: /opt/fivenines already exists or /home/fivenines does not exists"
+                exit 1
+        fi
+        sudo usermod -m -d /opt/fivenines fivenines
+
+        # Reinstall the agent due to path changes
+        echo "Reinstalling the fivenines_agent package"
+        sudo su - fivenines -s /bin/bash -c 'python3 -m pipx uninstall fivenines_agent'
+        sudo su - fivenines -s /bin/bash -c 'python3 -m pipx install --index-url https://pypi.org/simple/ --pip-args="--extra-index-url https://pypi.org/simple" fivenines_agent'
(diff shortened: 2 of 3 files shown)


fivenines_setup.sh
```diff
@@ -29,7 +29,7 @@ echo -n "$1" | sudo tee /etc/fivenines_agent/TOKEN > /dev/null
 
 # Create a system user for the agent
 if ! id -u fivenines >/dev/null 2>&1; then
-  sudo useradd --system --user-group --key USERGROUPS_ENAB=yes fivenines --shell /bin/false --create-home
```
spuyet, line 32: We also need to fix the [fivenines_update.sh](https://github.com/Five-Nines-io/fivenines_agent/blob/main/fivenines_update.sh) script to delete the old folder and reinstall in the right place if necessary.

bluet: for the transition for current users?

ok a moment

bluet: also patched update script and service file

fivenines_update.sh
```diff
@@ -1,4 +1,45 @@
 #!/bin/bash
+# This script is used to update the fivenines agent, and run by root
```
spuyet:
```suggestion
# This script is used to update the fivenines agent
```

fivenines_update.sh
```diff
@@ -1,4 +1,45 @@
...
+        echo "Migrating fivenines.io's working directory from /home/fivenines to /opt/fivenines"
+        # if /opt/fivenines exists, or /home/fivenines not exists, exit
+        if [ -d /opt/fivenines ] || [ ! -d /home/fivenines ]; then
+                echo "Error: /opt/fivenines already exists or /home/fivenines does not exists"
+                exit 1
+        fi
+        sudo usermod -m -d /opt/fivenines fivenines
+
+        # Reinstall the agent due to path changes
+        echo "Reinstalling the fivenines_agent package"
```
spuyet, line 19: Do we really need that step as the home folder has been moved ?

fivenines_update.sh
```diff
@@ -1,4 +1,45 @@
...
+        # Get the exit status of the pipx command
+        if [ $? -ne 0 ]; then
+                echo "Failed to Reinstall the fivenines_agent package"
+                exit 1
+        else
+                echo "Successfully Reinstalled the fivenines_agent package"
+        fi
+fi
+if [ "$(getent passwd fivenines | cut -d: -f6)" == "/opt/fivenines" ] && [ -d /opt/fivenines ]; then
+    echo "fivenines.io's working directory is already migrated to /opt/fivenines"
```
spuyet: This will always be printed as this is done after the migration, no ? Not sure that we need that step.

spuyet (commented):
Thanks for this PR, I will give it a try tomorrow :)

fivenines_update.sh
bluet: Yup. I tried to keep everything, but the pipx and modules saved the previous location as path references.

fivenines_update.sh
bluet: good point

bluet:
> Thanks for this PR, I will give it a try tomorrow :)

remember to manually patch service file when you try, as the PR (and fixes) has not been merged yet :smile:

spuyet:
@bluet Works like a charm thank you 🙏

Issue #4, opened 2024-12-04
FTBFS: required python version conflicts


bluet:
When trying to make a test build for #1 , it Failed To Build From Source ().

I see there's a poetry.lock file, so I assume I should use Poetry, right?

Python version:
```
$ python --version
Python 3.10.12
```
Poetry version: 1.8.4

Error logs:
```
$ poetry install --no-interaction
Creating virtualenv fivenines-agent-nYMUktQ2-py3.10 in /home/bluet/.cache/pypoetry/virtualenvs
Installing dependencies from lock file

pyproject.toml changed significantly since poetry.lock was last generated. Run `poetry lock [--no-update]` to fix the lock file.
```
```
$ poetry lock
Updating dependencies
Resolving dependencies... (0.5s)

The current project's supported Python range (>=3.5,<4.0) is not compatible with some of the required packages Python requirement:
  - psutil requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, so it will not be satisfied for Python >=3.5,<3.6.dev0
  
Because psutil (5.9.8) requires Python >=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*
 and no versions of psutil match >5.9.8,<6.0.0, psutil is forbidden.
So, because fivenines-agent depends on psutil (^5.9.8), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For psutil, a possible solution would be to set the `python` property to ">=3.6.dev0,<4.0"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```

Source from github main branch, latest v1.0.5 commit:
```
$ git status 
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
$ git pull
Already up to date.
$ git log -n 1
commit f035af25ead9d8ed911b1961f11b15a9005f9b9b (HEAD -> main, tag: v.1.0.5, origin/main, origin/HEAD)
Author: Sébastien Puyet <sebastienpuyet@gmail.com>
Date:   Tue Dec 3 16:54:44 2024 +0100

    v1.0.5
$ git remote -v
origin  https://github.com/Five-Nines-io/fivenines_agent.git (fetch)
[...]

PR #5, opened 2024-12-04
bump minimum python version to 3.6 to solve FTBFS #4

bluet:
bump minimum python version to 3.6 to solve FTBFS #4 due to mismatch with dependency (psutil)


spuyet:
Thank you!

Issue #1, opened 2024-12-03
Too many dependencies required during installation, even GCC is required.


bluet:
In current version, all python modules are installed by poetry, which requires gcc to be installed.

```
The following NEW packages will be installed:
  binutils binutils-common binutils-x86-64-linux-gnu cpp cpp-11 fonts-font-awesome fonts-lato gcc gcc-11 gcc-11-base javascript-common libasan6
  libatomic1 libbinutils libc-dev-bin libc-devtools libc6-dev libcc1-0 libcrypt-dev libctf-nobfd0 libctf0 libgcc-11-dev libgomp1 libisl23 libitm1
  libjs-bootstrap4 libjs-highlight.js libjs-jquery libjs-lunr libjs-modernizr libjs-popper.js libjs-sizzle liblsan0 libmpc3 libnsl-dev libquadmath0
  libtirpc-dev libtsan0 libubsan1 linux-libc-dev manpages-dev mkdocs node-jquery pipx python-babel-localedata python3-argcomplete python3-babel
  python3-jinja2 python3-livereload python3-markdown python3-markupsafe python3-pip-whl python3-psutil python3-pygments python3-setuptools-whl
  python3-tornado python3-userpath python3-venv python3.10-venv rpcsvc-proto sphinx-rtd-theme-common
```

Maybe use PyInstaller https://pyinstaller.org/en/stable/ to build self-contained distributable executables in CI/CD (GitHub Actions) as release file.

bluet:
A sample script here https://github.com/bluet/proxybroker2/blob/master/.github/workflows/py-to-exe.yml

spuyet:
That's definitely a good idea, I'm a bit worried about the binary size but we should give it a try.

bluet:
in the test in #9 it's just ~7MB. But I didn't test it yet.

```
$ ls -alh dist/linux/fivenines-agent-linux-amd64 
-rwxr-xr-x 1 bluet bluet 6.7M Dec  5 01:38 dist/linux/fivenines-agent-linux-amd64
```

bluet:
Maybe we should reopen the issue. 😅

Currently we only have the build script ready, but still need to work on automation, and migrate the installation process. 🚧🗿

spuyet:
@bluet I just released the `1.0.6` version as a single binary, first observations are:
- Binary file is big: ~15MB
- Installation is way faster, like less than 5 seconds (depending of the internet connection ofc)
[...]

PR #9, opened 2024-12-04
scripts for single-file build with pyinstaller for #1

bluet:
scripts for single-file build with pyinstaller for #1


bluet:
@spuyet with this PR, you can run `py2exe.sh` to generate an one-single-file release.

Currently only tested on Ubuntu 22.04. But if it works for you, I/we can implement a GitHub action to auto-build for multi-platform later in another PR.

spuyet:
@bluet first of all, thanks for what you're doing, I truly appreciate your help ☺️

I just gave it a try on a mac silicon, I had two issues:
```
pyinstaller ModuleNotFoundError: No module named 'distlib'
```
`distlib` seems to be necessary to activate a virtual env and is not automatically installed when I do `brew install pyenv-virtualenv`. I had to install it manually (`pip install distlib` but `pip install virtualenv` is probably cleaner).

```
./dist/linux/fivenines-agent-linux-amd64 
Traceback (most recent call last):
  File "py2exe_entrypoint.py", line 1, in <module>
    from fivenines_agent.agent import Agent
  File "PyInstaller/loader/pyimod02_importers.py", line 384, in exec_module
  File "fivenines_agent/agent.py", line 14, in <module>
    from fivenines_agent.ip import get_ip
  File "PyInstaller/loader/pyimod02_importers.py", line 384, in exec_module
  File "fivenines_agent/ip.py", line 4, in <module>
    import requests.packages.urllib3.util.connection
ModuleNotFoundError: No module named 'requests'
[PYI-69714:ERROR] Failed to execute script 'py2exe_entrypoint' due to unhandled exception!
```

the `requests` module is dynamically imported here and pyinstaller is not able to resolve this dependency at compile time so we just need to replace [this line](https://github.com/Five-Nines-io/fivenines_agent/blob/main/fivenines_agent/ip.py#L4) with `import requests`. I'll remove `requests` dependency (in a different PR) and use `http.client` as we do to communicate with the API, there is no need to use a different HTTP client.

As expected the binary size is fucking huge but I guess people prefer that 😢 :
```
8.3M Dec  4 20:59 fivenines-agent-linux-amd64
```

Removing the `requests` dependency make it loose a bit of weight :) :
```
7.6M Dec  4 21:18 fivenines-agent-linux-amd64
```

Let's fix these two things and we should be good. If you want to automate build with Github Action, it would be super nice but I can handle it if you want, your call, no pressure.

bluet:
Cool.  
Welcome, I'm happy to help. 😸  That's the open source way. Hahaha.

1. virtualenv related:
In autobuild we don't have to worry about the virtualenv error I think, as the GitHub action plugins could take care of it for us.  
As I don't have a Apple Silicon machine, would you like to patch the py2exe.sh to add MacOSX support? Or we can just leave it later and focus on supporting Linux servers first. (if you have Linux servers)

2. About the requests module:
Would you like to make the change in your repo/main first? I can sync/merge to this PR later to get it fixed.  
As I don't have write permission, if I open another PR to fix that and wait for review/merge and x/y/z and sync, it might be a little bit complicated. Lol

3. File size:
7.x MB is super small!! (Compare to other softwares).  
And downloading a 10MB file takes only few seconds, but installing a bunch of system packages and them python modules would take minutes!  
It's a huge win! :D

4. GitHub action:
I'm just here to help. This is your repo so you're the boss hahaha.  
Can also just create a issue first, and see who will be available first. 😉

bluet:
I'll have class to teach today, so I might spend most of the time in the university, but I'll try to find a chance to check here.  
If you can finish the `requests` module fix before that, I'll sync it to this PR for you ASAP, so you can check the results.  
And after that we can merge this PR first and go on with other enhancements 🚀

spuyet:
> 1. virtualenv related:
> In autobuild we don't have to worry about the virtualenv error I think, as the GitHub action plugins could take care of it for us.
> As I don't have a Apple Silicon machine, would you like to patch the py2exe.sh to add MacOSX support? Or we can just leave it later and focus on supporting Linux servers first. (if you have Linux servers)

Let's forget about that now, I will handle it later.

> 2. About the requests module:
> Would you like to make the change in your repo/main first? I can sync/merge to this PR later to get it fixed.
> As I don't have write permission, if I open another PR to fix that and wait for review/merge and x/y/z and sync, it might be a little bit complicated. Lol
 
I just removed it on `main` ;)

> 3. File size:
> 7.x MB is super small!! (Compare to other softwares).
> And downloading a 10MB file takes only few seconds, but installing a bunch of system packages and them python modules would take minutes!
> It's a huge win! :D

Well, I'd not say small but it will be ok for now haha but you're definitely right about the setup time 😄 

> 4. GitHub action:
> I'm just here to help. This is your repo so you're the boss hahaha.
> Can also just create a issue first, and see who will be available first. 😉

👍

bluet:
@spuyet I had synced and also updated py2exe.sh a bit.

If it works on your side (the building process), you can merge this first (Y)

Note: Due to the lack of access to server-side source code and the knowledge of your server-side tech stack, I can't test the functionalities of the resulting binary. But that's the story of another issue ticket. lol

py2exe.sh
```diff
@@ -0,0 +1,73 @@
...
+poetry install --no-interaction
+# poetry add -G dev poetry-plugin-export
+# poetry add -G dev pyinstaller
+
+#
+# Export dependencies to requirements.txt for backward compatibility with the old build process
+#
+
+echo "Exporting dependencies to requirements.txt for backward compatibility with the old build process"
+poetry export --without-hashes -o requirements.txt
```
spuyet, line 45: Could you elaborate a bit about why do we need this line ? Is the `pyproject.toml` file not enough ?

spuyet: BTW, do we need to track this file BTW ?

spuyet: ok forget about that, chat GPT explained me why ;)

PR #24, opened 2026-02-23
Synology (DSM7) Support

apooley:
Builds a Synology 7.x package in tandem with other packages (using the same pipeline) to be installable on Synology NAS units running DSM 7.x. Is missing some features due to permissions restrictions for Synology, but fallback is graceful and doesn't impact main builds. See README for full details.


.claude/plans/CLAUDE.md
```diff
@@ -0,0 +1,7 @@
+<claude-mem-context>
+# Recent Activity
+
+<!-- This section is auto-generated by claude-mem. Edit content outside the tags. -->
+
+*No recent activity*
+</claude-mem-context>
```
spuyet, line 7: I don't think we need all these CLAUDE files right now. I do prefer having only one CLAUDE file at the root of the project so easier to replace if I switch to anything else.

.github/workflows/build-release.yml
```diff
@@ -313,14 +315,148 @@ jobs:
...
   # Pre-release for feature and release branches
   pre-release:
-    needs: [build-amd64, build-arm64, build-alpine-amd64, build-alpine-arm64]
+    needs: [build-amd64, build-arm64, build-alpine-amd64, build-alpine-arm64, build-synology-amd64, build-synology-arm64]
     runs-on: ubuntu-latest
     if: |
       github.event_name == 'push' &&
       !startsWith(github.ref, 'refs/tags/') &&
-      (startsWith(github.ref, 'refs/heads/release/') || startsWith(github.ref, 'refs/heads/feature/'))
+      (startsWith(github.ref, 'refs/heads/release/') || startsWith(github.ref, 'refs/heads/feature/') || startsWith(github.ref, 'refs/heads/synology'))
```
spuyet, line 459: We don't need this, if we want to build specific synology stuff, we can just prefix branch with `feature/` as we do for others env/features.

CLAUDE.md
```diff
@@ -87,25 +100,28 @@ Each metric collector is a separate module that exports functions to collect spe
...
 - **Security**: `fail2ban.py` (requires sudo fail2ban-client)
+- **Network/connectivity**: `ip.py` (public IPv4/IPv6 via ip.fivenines.io with 60s cache), `ping.py` (TCP latency)
+- **Security scanning**: `packages.py` (installed packages via dpkg/rpm/apk/pacman with hash-based delta sync)
 
 Collectors use the `@debug` decorator from `debug.py` to log execution time and results.
 
 ### Configuration
 
-- Agent reads `TOKEN` file from config directory (`/opt/fivenines` for system install, `~/.local/fivenines` for user install)
+- Agent reads `TOKEN` file from config directory (default `/etc/fivenines_agent`, overridable via `CONFIG_DIR`)
```
spuyet: Nice catch for the wrong system install path but we need to keep the info about user install:

```suggestion
- Agent reads `TOKEN` file from config directory (default `/etc/fivenines_agent`, `~/.local/fivenines` for user install, overridable via `CONFIG_DIR`)
```

Dockerfile.synology
```diff
@@ -0,0 +1,39 @@
...
+    done
+
+# Configure libcrypt path for PyInstaller compatibility
+RUN echo "/usr/local/lib" >> /etc/ld.so.conf.d/libcrypt.conf && \
+    ldconfig
+
+ENV BASH_ENV=/etc/profile
+ENV TARGET_ARCH="amd64"
+WORKDIR /workspace
+CMD ["bash"]
```
spuyet, line 39: Do we really need a dedicated Dockerfile, looks like there is nothing fancy here

py2exe_synology.sh
```diff
@@ -0,0 +1,283 @@
...
+
+poetry install --no-interaction || {
+    echo "Poetry installation failed. Exiting."
+    exit 1
+}
+
+# Remove libraries not needed for Synology (no libvirt, no systemd watchdog, no proxmoxer)
+echo "=== Removing Synology-incompatible dependencies ==="
+pip uninstall -y libvirt-python systemd-watchdog proxmoxer || true
+echo "Removed libvirt-python, systemd-watchdog, and proxmoxer"
```
spuyet, line 117: If lines 100 to 124 are the only specify to synology we can probably just add them to the existing script to avoid having too many build scripts, WDYT ?

spuyet (commented):
Looks nice but we can probably simplify it a bit the content to prevent adding too many files and merge with what we already have

.claude/plans/CLAUDE.md
apooley: I... honestly didn't realize it went hog-wild creating Claude files. Sorry :(

.github/workflows/build-release.yml
apooley: Noted!

apooley:
All valid/noted; will clean up and resubmit for you! (apologies, I'm normally a single-maintainer private repo builder, lol)

PR #25, opened 2026-02-23
Security Updates

apooley:
Ultimately this is largely unnecessary hardening, but basically a pass over the agent for "What if fivenines' API got compromised?" and hardening things appropriately. 

Note, I don't think it's likely to be an issue, but defence-in-depth and all that - figured it was best to do a pass and protect against abuse just in case. Would definitely lower your blast radius if anything *did* happen.


spuyet:
Could you cleanup all the synology related stuff from this branch please ? Would be easier to review ;)

apooley:
My bad, dude; I was in the wrong damn branch when I worked on it and didn't realize. Let me close it, simplify/clean it up, and resubmit.

PR #27, opened 2026-02-24
Security Hardening

apooley:
## Security: Harden agent against compromised API and local credential exposure

This PR adds validation and hardening so the agent is safe to run when the API backend may be compromised or when a local unprivileged user can read the config directory.

### Threat model

- **Compromised API:** Attacker controls the API and can send arbitrary config. Goal: prevent that config from redirecting collectors, stealing credentials, or running arbitrary commands.
- **Local unprivileged user:** Attacker on the same host. Goal: prevent exposure of the agent token via relaxed file permissions.

### Changes

**1. Central config validation (`config_schema.py`)**

- Every config dict from the API is passed through `validate_config()` before use.
- Whitelist-based: only known keys are kept; each is validated and sanitized.
- Invalid or malicious values disable the affected collector only; the rest keep running.

**Collector constraints enforced:**

| Collector   | Constraint |
|------------|------------|
| nginx, caddy | Status/admin URLs restricted to loopback only (no SSRF to internal or cloud-metadata endpoints). |
| postgresql | Host restricted to loopback; CRLF stripped from password. |
| proxmox    | Host loopback-only; `verify_ssl` forced to `True`. |
| docker     | `socket_url` must be `None` or `unix://` (no TCP/HTTP). |
| qemu       | URI allowlisted to `qemu:///system` or `qemu:///session` only. |
| redis      | Password CRLF stripped; port range 1–65535. |
| ports      | Monitored ports validated as integers in range. |

Interval and request options (timeout, retries, etc.) are clamped to safe ranges.

**2. Redis CRLF injection (FN-SEC-001)**

- Redis collector switched from inline protocol to RESP; arguments are length-prefixed so CRLF in the password cannot inject commands.
- Redis password is also sanitized at config ingestion (CRLF stripped).

**3. Token file permissions (FN-SEC-002)**

- When writing the per-host token, the agent now uses `os.open(..., 0o600)` so the TOKEN file is always created with mode `0600`, independent of process umask.

**4. Proxmox and QEMU**

- Proxmox and QEMU collectors now receive only validated config from `validate_config()` (loopback host, forced SSL verification for Proxmox; allowlisted URI for QEMU).

**5. Documentation**

- **SECURITY.md** — Threat model, transport security, config validation, credential storage, and design trade-offs.
- **FN-Security-Findings.md** — Detailed findings, remediation status, and proof-of-concept notes (for reference; not merged into main docs).
- **README.md** — New “Security” section with a short summary and link to SECURITY.md.

### Testing

- New tests for `config_schema` (validation and sanitization for all constrained collectors).
- New/updated tests for Redis RESP encoding and CRLF-safe behavior.
- New tests for token file permissions and synchronizer config handling.
- New tests for Proxmox and QEMU when given invalid or remote config (collectors disabled).

All existing and new tests pass.


spuyet:
Thanks for the PR, I just created another based on yours to add/update/remove some critical changes: @https://github.com/Five-Nines-io/fivenines_agent/pull/28

I need to test it carefully before merging into main as it could break existing installation.

PR #29, opened 2026-03-11
SELinux support tested with RockyLinux 9

istheway:
* fivenines_setup.sh modified to install selinux/fivenines_agent.te and selinux/fivenines_fc

### DIFF
--- a/fivenines_setup.sh
+++ b/fivenines_setup.sh
@@ -122,23 +122,132 @@ detect_system() {
   fi
 }
 
+
+# Check if SELinux is installed and configure contexts if Enforcing
+setup_selinux_contexts() {
+  if ! command -v sestatus >/dev/null 2>&1; then
+    print_success "SELinux is not installed on this system."
+    return 0
+  fi
+
+  # Parse SELinux mode using sestatus
+  selinux_mode=$(sestatus 2>/dev/null | awk -F: '/^Current mode:/ {gsub(/^[ \t]+|[ \t]+$/, "", $2); print $2}')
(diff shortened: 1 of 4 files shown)


spuyet:
@istheway looks nice, thank you. We also need to update https://github.com/Five-Nines-io/fivenines_agent/blob/main/fivenines_update.sh and https://github.com/Five-Nines-io/fivenines_agent/blob/main/fivenines_uninstall.sh

Not sure if the user only version (https://github.com/Five-Nines-io/fivenines_agent/blob/main/fivenines_setup_user.sh) would work without tweaking some selinux rules, I need to give it a try.

istheway:
@spuyet I tested it on RockyLinux 9, AlmaLinux 8, 9 and 10 and the dashboard reported correctly as far as I could see in the UI.  You may want to test it with fedora and rhel.

I have not tested the update sh but it probably does not matter since contexts are already installed upon an update, they should not change...   As to the uninstaller, cleaning off the contexts is a good idea when running the uninstall.

istheway:
You need sudo root access for the selinux context module installation so would not be appropriate for user only shell installation for selinux enforced systems.

istheway:
I updated the fivenines_uninstall.sh so that the selinux contexts modules get removed.  Let me know how your tests go.  I tested on Centos 9 and 10 as welll during this process.

Suggestion: agent poll update should probably happen every 5 seconds by default, or every 1 second as option.

spuyet:
Thanks for your PR @istheway I've extended it added CI integration, it's now merged into main.

PR #46, opened 2026-05-09
Fix: cache get_ip failures to stop journal spam (closes #42)

### LINKED ISSUE #42: ip.py floods journal with errors when host has no IPv6 connectivity
## Symptom

On hosts without IPv6 connectivity (or with broken IPv6 routing), the agent retries `get_ip(ipv6=True)` every tick. Each failure produces 2 error lines in the systemd journal:

```
fivenines_agent[50542]: [ERROR][thread#XXXXX] Unexpected error occurred: Could not connect to ip.fivenines.io on port 443 with family IPv6
fivenines_agent[50542]: [ERROR][thread#XXXXX] Could not connect to <ipv6>: [Errno 101] Network is unreachable
```

At default 60s tick interval, that's ~2880 lines/day per failing host. Multiplied across regions/IPv6 endpoints in `request_options`, can exceed 5000+ lines/day.

## Source

- `fivenines_agent/ip.py:79-87` — `ConnectionError` and generic `Exception` are both logged at `error` level and the function returns `None`. No backoff, no caching of the failure, no per-host suppression.
- `fivenines_agent/ip.py:54-58` — there's already a 60s positive-result cache, but no negative-result cache.
- `fivenines_agent/agent.py:154-157` — agent calls `get_ip(ipv4=...)` / `get_ip(ipv6=...)` every tick when `config["ipv4"]` / `config["ipv6"]` is truthy.

## Why this matters

1. Operator log noise: legitimate errors get buried under known-broken-IPv6 spam.
2. Disk/journal pressure on storage-constrained hosts (Proxmox, container hosts).
3. The agent already does the right thing semantically (returns `None`, sends `ipv6: null` to the backend) — the noise is purely a logging/retry policy issue.

## Suggested approaches

Three options, ordered by complexity:

1. **Negative-result cache** (smallest): cache the failure for N seconds (e.g., 5min) the same way the success path is cached at lines 54-58. After a failure, return the cached `None` without retrying. Re-attempt only after the TTL.
2. **Detect IPv6 stack absence at startup**: when the kernel routing table has no IPv6 default route, set the `ipv6` capability to False in `permissions.py` (similar pattern to other capability probes). The agent then never calls `get_ip(ipv6=True)`. Solves the root cause but ties IPv6 to the capability system.
3. **Exponential backoff with rate-limited error log**: log the first error normally, then suppress subsequent identical errors for an increasing window. Standard Python logging idiom.

Recommendation: **#1** (negative cache) — symmetrical with the existing 60s positive cache, minimal diff, no new abstractions, also reduces wasted network attempts on the unhappy path.

## Reproduction

Run the agent on a host with no IPv6 default route (Proxmox / container / IPv4-only network) with `ipv4: true, ipv6: true` in config. Tail journal:

```
journalctl -u fivenines-agent -f | grep -c "family IPv6"
```

Watch the count climb at 4/min.

## Note

Surfaced during the eng review for the GPU silent-failure fix (`spuyet/gpu-debug-logging`). Not related to that PR — separate noise issue.

spuyet:
## Summary

Closes #42.

On hosts with broken or absent IPv6 connectivity, `get_ip(ipv6=True)` fails on every tick. Each failure produces 2 error lines in the systemd journal (the inner `Could not connect to <ip>` from `CustomHTTPSConnection.connect` and the outer `Unexpected error occurred` from `get_ip`). At default 60s tick, that's ~5760 lines/day for a single broken host. Surfaced during the GPU silent-failure investigation (#43).

The agent had a 60s positive cache (`ip.py:54-58`) for successful fetches, but no corresponding negative cache. Every failure path through lines 79-87 re-attempted the network call and re-logged the same error.

## What changes

`fivenines_agent/ip.py`:

- New module constant `NEGATIVE_CACHE_TTL = 300` (5 minutes -- longer than the 60s positive cache and longer than the default tick interval, so a broken host doesn't flap into the cache window every minute).
- Every failure path (HTTP non-200, `ConnectionError`, generic `Exception`) now updates the cache: `ip=None, timestamp=now`. Subsequent calls within 300s short-circuit at the new check and return `None` silently.
- Cache check refactored into one block at the top of `get_ip`:
  - Positive hit (`ip is not None and age < 60s`) returns the cached IP.
  - Negative hit (`ip is None and timestamp > 0 and age < 300s`) returns `None` silently.
  - Otherwise falls through to a fresh fetch.
- `conn = None` initialized before the `try` block to keep the `finally` clause correct on early exits (was a latent NameError if `ssl.create_default_context` ever raised).

## Operator-visible result on the diagnosed host

Before:
```
[ERROR] Could not connect to 2a01:4f8:c013:ad01::1: [Errno 101] Network is unreachable
[ERROR] Unexpected error occurred: Could not connect to ip.fivenines.io on port 443 with family IPv6
[ERROR] Could not connect to 2a01:4f8:c013:ad01::1: [Errno 101] Network is unreachable
[ERROR] Unexpected error occurred: Could not connect to ip.fivenines.io on port 443 with family IPv6
... (every 60 seconds, forever)
```

After:
```
[ERROR] Could not connect to 2a01:4f8:c013:ad01::1: [Errno 101] Network is unreachable
[ERROR] Unexpected error occurred: Could not connect to ip.fivenines.io on port 443 with family IPv6
... (silence for 5 minutes)
[ERROR] Could not connect to 2a01:4f8:c013:ad01::1: [Errno 101] Network is unreachable
[ERROR] Unexpected error occurred: Could not connect to ip.fivenines.io on port 443 with family IPv6
... (silence for 5 minutes)
```

That's a 60x reduction in error volume per failed host, while still leaving a periodic heartbeat so the operator knows the failure mode is current.

## Recovery

Recovery is automatic: once the 300s negative TTL expires, the next call re-attempts. On success, the cache transitions cleanly from `(ip=None)` to `(ip=<new IP>)` and the next 60s of calls return the cached IP.

## Test plan

- [x] 376 tests pass (7 new in `test_ip.py`)
- [x] Existing 8 tests still pass (positive cache, expiration, all three failure modes)
- [x] New tests cover:
  - `ConnectionError` populates negative cache; 60 subsequent calls produce zero log entries and zero HTTP attempts
  - Generic `Exception` populates negative cache
  - HTTP non-200 populates negative cache
  - `NEGATIVE_CACHE_TTL` expiration triggers a fresh attempt
  - IPv6 failure does not poison IPv4 cache (and vice versa)
  - Recovery: after TTL expires, success replaces cached `None` with the new IP

## Out of scope

- Detecting IPv6 stack absence at startup and gating `get_ip(ipv6=True)` via the capability system. Considered in #42 as option #2 -- defer until needed.
- Exponential backoff. Considered in #42 as option #3 -- the symmetric cache is simpler and matches the existing positive-cache pattern.

### DIFF
--- a/fivenines_agent/ip.py
+++ b/fivenines_agent/ip.py
@@ -65,11 +65,14 @@ def _is_public_ip(parsed):
 def _validate_ip(body, ipv6):
     """Return body if it parses as a public address of the expected family.
 
-    Length is checked on the raw body BEFORE stripping, so a hostile or
-    broken upstream cannot smuggle a valid-looking IP inside whitespace
-    padding ("1.2.3.4" + " " * 500 strips to a valid IP).
+    Length is checked on the encoded BYTE length (not character count) so
+    multibyte Unicode whitespace cannot smuggle a valid-looking IP through
+    .strip() ("1.2.3.4" + " " * 57 is 64 chars but 178 bytes). The
+    check happens before .strip() so trailing/leading padding is also caught.
     """
-    if not body or len(body) > MAX_RESPONSE_BODY:
+    if not body:
+        return None
+    if len(body.encode("utf-8", errors="replace")) > MAX_RESPONSE_BODY:
         return None
     candidate = body.strip()
     if not candidate:
@@ -157,9 +160,18 @@ def get_ip(ipv6=False):
         conn.request("GET", "")
         response = conn.getresponse()
         # Cap the read so a misbehaving (or hostile) upstream cannot force
-        # unbounded memory use or log amplification. Strict UTF-8 decode so
-        # non-UTF-8 bytes are not silently dropped to forge a valid-looking IP.
+        # unbounded memory use or log amplification. Reading past the cap is
+        # how we detect oversized bodies (otherwise a 1MB body would just look
+        # like a 64-byte body via silent truncation).
         raw = response.read(MAX_RESPONSE_BODY * 4)
+        if len(raw) > MAX_RESPONSE_BODY:
+            log(
+                f"ip.fivenines.io returned oversized body ({len(raw)} bytes)",
+                "error",
+            )
+            _record_failure(cache)
+            return None
+        # Strict UTF-8 decode so non-UTF-8 bytes are not silently dropped.
         try:
             body = raw.decode("utf-8")
         except UnicodeDecodeError:
--- a/fivenines_agent/ip.py
+++ b/fivenines_agent/ip.py
@@ -11,6 +11,12 @@ from fivenines_agent.debug import debug, log
 _ip_v4_cache = { "timestamp": 0, "ip": None }
 _ip_v6_cache = { "timestamp": 0, "ip": None }
 
+# Cache TTLs. The negative cache deliberately runs longer than the positive
+# cache (and longer than the default tick interval) so a host with no IPv6
+# does not log the same connection error every tick. See issue #42.
+POSITIVE_CACHE_TTL = 60
+NEGATIVE_CACHE_TTL = 300
+
 class CustomHTTPSConnection(http.client.HTTPSConnection):
     def __init__(self, host, port=None, ipv6=False, timeout=5, **kwargs):
         super().__init__(host, port, timeout=timeout, **kwargs)
@@ -49,14 +55,17 @@ class CustomHTTPSConnection(http.client.HTTPSConnection):
 @debug('get_ip')
 def get_ip(ipv6=False):
     now = time.time()
+    cache = _ip_v6_cache if ipv6 else _ip_v4_cache
+    age = now - cache["timestamp"]
 
-    if ipv6:
-        if now - _ip_v6_cache["timestamp"] < 60:
-            return _ip_v6_cache["ip"]
-    else:
-        if now - _ip_v4_cache["timestamp"] < 60:
-            return _ip_v4_cache["ip"]
+    # Positive cache: short TTL, return the previously-fetched IP.
+    if cache["ip"] is not None and age < POSITIVE_CACHE_TTL:
+        return cache["ip"]
+    # Negative cache: longer TTL, suppress retry storms on broken connectivity.
+    if cache["ip"] is None and cache["timestamp"] > 0 and age < NEGATIVE_CACHE_TTL:
+        return None
 
+    conn = None
     try:
         ssl_context = ssl.create_default_context(cafile=certifi.where())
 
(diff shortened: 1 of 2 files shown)


spuyet:
Updated after `/codex challenge` — the original flat 300s cache had a real regression Codex caught:

> One transient timeout / dropped packet / brief ip.fivenines.io outage after cache expiry turns into five minutes of false "no public IP" data. On `origin/main`, the same failure only affected that sample and retried on the next tick. This is a silent data-corruption path.

Replaced with exponential backoff:

| Consecutive failures | Negative cache TTL |
|---|---|
| 1 | **0** (no cache, retries next tick) |
| 2 | 60s |
| 3 | 120s |
| 4 | 240s |
| 5+ | 300s (cap) |

Single transients now match main behavior. Permanent failures still reach the 5-minute cap. Any successful fetch resets the counter.

Tests updated: 18 in `test_ip.py` (was 15). The first failure no longer caches (test reflects this), and there's a new schedule-verification test, a per-family isolation test, and a 60-tick steady-state quiet check that exercises the actual backoff curve.

Quantitative impact for the diagnosed Proxmox host:

| Scenario | Before #46 | Flat 300s (initial commit) | Exponential backoff (now) |
|---|---|---|---|
| Permanent IPv6 break | 5760 lines/day | 576/day | ~600/day |
| Single transient blip | 2 lines + recovery next tick | 2 lines + 5 min of null | 2 lines + recovery next tick |
| 3-tick brief outage | 6 lines + recovery | 2 lines + ~5 min of null | 4 lines + recovery in ~1 min |

spuyet:
Round 2 of `/codex challenge` confirmed the round-1 findings are fixed and surfaced 3 more, all of which I've now addressed:

| Round 2 finding | Severity | Status |
|---|---|---|
| `now` captured at function entry, then used for failure timestamp (slow failures have shorter-than-advertised suppression) | HIGH | Fixed: `_record_failure()` calls `time.monotonic()` at write time |
| 200 response with garbage body (empty / HTML / wrong family) cached as "the IP" and reset the failure counter | HIGH | Fixed: new `_validate_ip()` rejects via `ipaddress.ip_address()`, family check, max-length check; failures are counted, not reset |
| Wall-clock for age math (NTP step, suspend, VM restore could keep cache suppressed past the 300s cap) | MEDIUM | Fixed: switched to `time.monotonic()` |
| Cached failures invisible to telemetry (same finding as round 1) | MEDIUM | Deferred: requires payload-format changes beyond this PR |

12 new tests:

- `_validate_ip()` helper: 8 cases (valid v4, valid v6, empty, HTML, garbage, wrong family v4-for-v6, wrong family v6-for-v4, oversized)
- `get_ip()` integration: invalid 200 body counts as failure, wrong family counts as failure, garbage 200 does NOT reset an existing failure streak
- Clock-rollback resilience: forcing the timestamp far into the future (simulating a backward wall-clock jump) does not keep the cache suppressed past the cap

391 tests pass total. ASCII clean. Black + isort clean.

PR #67, opened 2026-06-04
feat: backend-controlled permission re-check with reliable QEMU probe

spuyet:
Makes the agent detect late-arriving capabilities (QEMU, Docker, SMART, ...) within ~one interval instead of up to 5 minutes, so new users stop seeing an empty dashboard and thinking the integration is broken, without ever coupling collection cadence to permission state.

- Selective gap re-probe of enabled-but-missing capabilities every tick (the 300s full probe is kept for regressions).
- QEMU is now probed via `libvirt.openReadOnly("qemu:///system")` behind a hard timeout (worker thread + join), mirroring the collector exactly so a wedged libvirt stack can no longer stall the whole loop; the probe spec is refactored into one shared `name -> (probe, args)` table.
- Backend-tunable recheck throttle (floored at the collection interval) and an on-demand re-detect token, with `pending_capabilities` reported in the payload for the dashboard "pending + reason" view, plus a collection-interval clamp.
- 39 new tests (`test_permissions_recheck.py`, `test_agent_recheck.py`); full suite green (583 passing).

Pre-merge gates (cannot run on the macOS dev box): verify `openReadOnly` on a real libvirt host (system + modular `virtqemud`/`virtproxyd`) and run `make test` on Linux, since CI does not run pytest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,3 +162,4 @@ cython_debug/
 /TOKEN
 
 .install.stamp
+.gstack/
--- a/fivenines_agent/agent.py
+++ b/fivenines_agent/agent.py
@@ -17,7 +17,12 @@ except ImportError:
     systemd_watchdog = None
 
 from fivenines_agent.cli import VERSION
-from fivenines_agent.collectors import _collect_with_telemetry, collect_metrics
+from fivenines_agent.collectors import (
+    COLLECTORS,
+    _capability_key_for,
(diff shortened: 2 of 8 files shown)


spuyet:
Superseded by #68 — branch renamed to `feature/spuyet/backend-controlled-permission-recheck` so the CI builds a pre-release. Same commits.

PR #69, opened 2026-06-14
feat: connect to PostgreSQL via pg8000 driver instead of psql

spuyet:
Replaces the PostgreSQL collector's `psql` shell-out with a direct pg8000 (pure-Python) socket connection, so Postgres running inside a Docker container is monitored with no `psql` binary on the host. The collector now returns a reachability-aware payload (`reachable`/`error` categories, with per-query partial success preserved once connected) and bounds each query with a connect + statement timeout so a slow query can't stall the collection loop; a path-like `host` is routed to a unix socket so existing socket-directory configs keep working. pg8000 (plus scramp + python-dateutil) is added to the deps and wired into the PyInstaller builds (glibc/musl/Windows) via hidden-imports. Includes a new 100%-coverage unit suite and an opt-in integration test (`RUN_PG_INTEGRATION`) that exercises a real socket and SCRAM auth.

Note: the new `reachable`/`error` payload is a `/collect` contract change; the backend should accept it before the agent's emit is relied on (backward-compatible spec handed off separately). Multi-instance Postgres support is intentionally out of scope.

### DIFF
--- a/fivenines_agent/postgresql.py
+++ b/fivenines_agent/postgresql.py
@@ -1,5 +1,6 @@
 import os
 import socket
+import ssl
 
 import pg8000.dbapi
 from pg8000.exceptions import DatabaseError, InterfaceError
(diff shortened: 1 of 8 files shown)


spuyet:
Superseded by #70. Same change, moved to `feature/spuyet/baghdad-v2` so the CI build/release workflow triggers (the `feature/**` branch filter) and the artifacts can be tested live. Continue review on #70.

PR #93, opened 2026-07-23
v1.12.0 feat: MQTT collector: persistent broker subscription + per-topic state reporting (#92)

### LINKED ISSUE #92: MQTT collector: persistent broker subscription + per-topic state reporting
_Agent half of Five-Nines-io/fivenines_server#608 (MQTT monitoring: org-level brokers + topic freshness / payload checks)._

## Summary
Subscribe to customer MQTT brokers (config-pushed by the server on `/collect`, same delivery seam as `snmp_targets`) with a **persistent background client**, maintain in-memory per-topic state, and report a bounded snapshot every tick under `data["mqtt"]`. This is the agent's first long-lived-connection subsystem - the design deliberately leans on paho-mqtt's own network thread to absorb that novelty.

## Config contract (echoed on every /collect, replaces agent config wholesale)
```json
"mqtt": [
  {
    "broker_id": "…", "host": "…", "port": 8883, "tls": true,
    "username": "…", "password": "…",
    "monitors": [
      {"id": "…", "topic_filter": "iot/+/status", "capture_payload": true}
    ]
  }
]
```
- Key absent / `false` = tear down all MQTT clients.
- Diff desired-vs-current per tick: start/stop/resubscribe only on change (connection churn is not free on brokers; a wholesale reconnect every tick would also permanently re-trigger retained replays).

## Client (proposal: paho-mqtt - needs maintainer sign-off on the dependency)
- **paho-mqtt**: pure Python, no C extension, the canonical client. `loop_start()` background network thread, built-in reconnect/backoff, and `on_message` exposes `msg.retain` - the flag the whole feature's honesty hangs on.
- Alternative considered and rejected: `mosquitto_sub` subprocess (the mysql CLI shell-out posture) - fragile persistent-stream management, external binary requirement on every distro, retain flag only via `-F '%R …'` formatting. A persistent subscription is a poor fit for subprocess supervision.
- Protocol scope v1: MQTT 3.1.1, TCP + TLS (`tls_set`), username/password auth. Subscriptions at **QoS 0 with clean session**: QoS 1 + persistent sessions would deliver queued messages late after a reconnect, poisoning freshness ages with no way to date them - the arming rule (server side, fivenines_server#608 rule 3) covers the gap honestly.
- Out of scope v1: MQTT 5, WebSockets transport, mTLS client certs, publishing (the agent never publishes).

## Per-topic state (in-memory, snapshotted per tick)
Reported grouped **by monitor id** (so the server never re-implements wildcard topic matching), keyed by concrete topic:
- `last_live_seen_age_s` - updated ONLY by RETAIN=0 deliveries
- `last_message_age_s` - any delivery
- `last_payload` (truncated <= 1 KB, only when `capture_payload`), `last_payload_retained`
- `first_seen_age_s`

Plus per-broker connection envelope: `status` (connected / error / auth_error), `error` detail, `connected_age_s`, and per-monitor `subscribed_age_s` (the server's alarm-arming input). **Ages, not timestamps** - the server anchors everything to its own `received_at`, so agent clock drift is irrelevant.

## Honesty + bounds (contract-critical)
- RETAIN=1 deliveries must NEVER update `last_live_seen_age_s` (a stored status replay is not device activity). This is the Uptime Kuma flaw we are refusing to ship.
- Collection failure (client dead, connect/auth error, config not yet applied) -> report the error envelope; never fabricate topic entries; never report an empty-but-healthy shape on failure (the docker null contract: absence must be distinguishable from emptiness).
- Auth failure -> `auth_error` (maps to the server's amber `config_error`, the mysql envelope pattern).
- Caps: <= 500 concrete topics per monitor (stop discovering, set a `capped` flag), payload truncation before buffering, non-UTF-8 payloads carried safely (replace/scrub before JSON encoding).
- Clean teardown on config removal and agent shutdown (`loop_stop` + `disconnect`), no thread leaks across config reloads.

## Contract fixture
`tests/fixtures/mqtt_contract_payload.json` - **source of truth here**, vendored byte-identical into the server repo (the ceph/zfs/docker discipline). Scenarios to pin: healthy multi-topic, retained-only topic (fresh payload state, no live freshness), wildcard fan-out, auth failure, broker unreachable, capped discovery.

## Acceptance criteria
- [ ] Tick-driven client lifecycle (create / reconfigure / teardown / reconnect) covered by tests (mocked paho callbacks or a local broker in CI).
- [ ] Retained-vs-live distinction pinned by tests on the exact `msg.retain` path.
- [ ] Bounded memory under a topic storm (cap enforced + `capped` flag reported).
- [ ] `data["mqtt"]` absent when the feature is unconfigured (old servers unaffected); unknown config keys ignored (old agents unaffected).
- [ ] Ships in the next minor - the server pins `FEATURES_SUPPORTED_VERSIONS["mqtt"]` to that release. Deploy order: server first, agent release after.

**Effort:** M. Pattern: `snmp_targets` config consumption; new ground: the persistent client thread (delegated to paho).

spuyet:
Agent half of Five-Nines-io/fivenines_server#608. Closes #92.

## Summary

Adds the agent's **first long-lived-connection subsystem**: a persistent `paho-mqtt` 2.x client per broker (its own `loop_start` network thread) that subscribes to customer MQTT brokers, maintains in-memory per-topic freshness state, and reports a bounded snapshot every tick under `data["mqtt"]`.

Config is pushed by the server on `/collect` under the top-level `mqtt` key, the same delivery seam as `snmp_targets` (a special-case collector, not the `COLLECTORS` registry — no capability gate, paho is bundled). Each tick an `MQTTManager` diffs desired-vs-current and acts **only on change** (start / stop / resubscribe) — it never reconnects on an unchanged config, which would re-trigger retained-message replays every tick.

## Honesty contract (why this feature exists)

- **RETAIN=1 deliveries update `last_message_age_s` but NEVER `last_live_seen_age_s`.** A stored status replay is not proof a device is alive — this is the Uptime-Kuma flaw we refuse to ship. Pinned by tests on the exact `msg.retain` path.
- **Absence vs failure are distinct.** `None` = unconfigured, so the agent omits `data["mqtt"]` entirely (old servers unaffected, key never appears as bare `null`). A broken broker returns an `error` / `auth_error` envelope — never `None`, never fabricated topics. `auth_error` maps to the server's amber `config_error`.
- **Ages, not timestamps.** The server anchors freshness to its own `received_at`, so agent clock drift is irrelevant; `subscribed_age_s` is the server's alarm-arming input.

## Scope (v1)

MQTT 3.1.1 over TCP/TLS, username/password auth, **QoS 0 + clean session**, subscribe-only. Caps: 500 topics/monitor (`capped` flag), 1 KB payload truncation, non-UTF-8 scrubbed with `errors="replace"`. Clean teardown on config removal and agent shutdown (`loop_stop` + `disconnect`); per-broker construction **and restart-rebuild** failures isolated so one bad broker never hides the others or reports a stale `connected`.

Deferred to TODOS (P3): MQTT 5, WebSockets, mTLS client certs, `tls_insecure`, publish-based liveness.

## Acceptance criteria

- [x] Tick-driven lifecycle (create / reconfigure / teardown / reconnect) covered by tests (mocked paho callbacks)
- [x] Retained-vs-live distinction pinned on the exact `msg.retain` path
- [x] Bounded memory under a topic storm (cap enforced + `capped` flag)
- [x] `data["mqtt"]` absent when unconfigured; unknown config keys ignored (malformed brokers/monitors skipped)
- [x] Ships next minor (**1.12.0**); server pins `FEATURES_SUPPORTED_VERSIONS["mqtt"]` to it

## Testing

- `fivenines_agent/mqtt.py` at **100% coverage** (66 unit tests)
- Full suite: **1517 passed, 2 skipped** (`make test` can't run natively on macOS due to libvirt; verified in a scoped venv — CI's Linux jobs build/lint only)
- Cross-repo contract pinned by `tests/fixtures/mqtt_contract_payload.json` (6 scenarios: healthy multi-topic, retained-only, wildcard fan-out, auth failure, broker unreachable, capped discovery) + a replay-driven round-trip test
- Independent adversarial pre-landing review found **no landing-blockers**; 2 low-severity honesty/doc fixes applied (stale-status-on-failed-rebuild, disabled-branch comment accuracy)

## Deploy order (lockstep, server-first)

Server first, agent release after. Pending in `fivenines_server#608`: the `Ingesters` MQTT path, `FEATURES_SUPPORTED_VERSIONS["mqtt"] = 1.12.0`, a byte-identical copy of the contract fixture, and the org-level broker config UI. The agent is inert until the backend sends `mqtt` config.


spuyet (commented):
Contract-conformance + adversarial review against the #92 pinned contract (fivenines_server#608 is waiting on this to start). Verdict at the bottom. Every load-bearing honesty rule checked out against the code:

- Retained-vs-live: `_record` updates `last_message_at` always and `last_live_at` only `if not retained` (mqtt.py:360-363), pinned by `test_retained_message_never_sets_live` + `test_later_retained_does_not_reset_prior_live`.
- Clean-session resubscribe on every (re)connect (`_on_connect` -> `_resubscribe_all`, mqtt.py:307-308) - the classic paho reconnect bug is handled.
- Arming resets on disconnect (`_on_disconnect` nulls `subscribed_at` + clears `_subscribed_filters`, mqtt.py:313-319), and topic state deliberately survives the drop, so a disconnected broker reports truthful last-observed ages under an `error` envelope.
- Thread safety: per-broker `_lock` on all mutable state, wire calls kept outside the lock, `_brokers` dict confined to the tick thread; the one interleaving I could construct (`_resubscribe_all` racing `update_monitors`) converges within one tick because `reconcile` re-applies monitors every tick.
- paho 2.x VERSION2 callback signatures hand-verified (the tests mock paho, so a mismatch would pass CI); `connect_async` + `loop_start` keeps a dead broker from stalling the tick loop; `tls_set()` = real verification, no insecure path; caps + truncation + malformed-config tolerance all covered by tests.

Three asks, none landing-blockers:

**1. [P2] Add a `disconnected_after_data` scenario to the contract fixture.** The fixture is what the server's e2e ingester spec replays, and its most safety-critical branch has no pinned bytes: the state where staleness alerts must be GATED (broker was connected, collected topics, then dropped). The behavior is implemented - `_on_disconnect` clears only `subscribed_at`/`_subscribed_filters`, `mon.topics` survives, and `snapshot` reports them under `"status": "error"` - and unit-tested (`test_on_disconnect_marks_error_and_disarms`), but the server needs the exact payload shape: `status: "error"`, `error: "disconnected: ..."`, `connected_age_s: null`, `subscribed_age_s: null`, topics present with growing ages. Suggested events: connect -> two live messages -> disconnect -> snapshot. Low effort, directly consumed by the backend build.

**2. [P2, contract doc] Pin the real absence semantics in the fixture description: "absent while configured = collection failure".** The `null = collector crashed` branch from the #92 contract comment is dead code in practice: `_collect_with_telemetry` catches every exception and returns `None` (`collectors.py:146` - `except Exception as e: ... return None`), so a crash and an unconfigured host both produce an ABSENT `data["mqtt"]`. That is fine (simpler, and the server can disambiguate because it knows what it configured), but the rule now lives entirely server-side and deserves one sentence in the fixture `description`, since the fixture is the cross-repo contract artifact. Same for the per-broker variant: `_start_broker` skips a broker whose construction raised (mqtt.py:545-549), so the `mqtt` key can be present with a configured `broker_id` missing - the server must treat that broker as collection-failed, never as removed. Mirroring both on fivenines_server#608.

**3. [P3] `_wire_subscribe` ignores paho's return code** (mqtt.py:388-392). `client.subscribe()` does not raise on a dropped connection - it returns `(MQTT_ERR_NO_CONN, None)` - so the `except` can never catch that case, and `subscribed_at` is still armed (`_resubscribe_all` arms after the wire calls, `update_monitors` at intent time). Self-healing within seconds (a real drop fires `_on_disconnect` -> disarm; the next connect resubscribes), but checking `rc` and skipping the arm makes `subscribed_age_s` strictly truthful for a 2-line change. Related known limitation worth a TODOS line alongside the existing MQTT block: no `on_subscribe` SUBACK verification - noting that Mosquitto deliberately answers success for ACL-denied subscriptions and silently drops deliveries, so client-side detection is structurally unreliable and the server's exact-topic staleness is the real safety net.

Accepted as-is, for the record: the payload shape diverges from the array-based proposal in the #92 comment (dict-keyed `brokers`/`monitors`/`topics`, envelope inlined) - agent repo is the contract source of truth and the map shape is better; the server builds against THIS fixture. The disabled-host-keeps-clients deferral is documented inline in agent.py and acceptable for v1.

Transparency note: the cross-model adversarial passes both failed on infrastructure this run (Codex 5-min timeout, Claude subagent session limit), so this is a single deep human+model pass, not the usual multi-model review. The mocked-paho signature risk was hand-verified against paho-mqtt 2.1 VERSION2; a real-broker smoke test remains a nice-to-have (CI is build/lint only).

**Verdict: good to merge after (or with) #1, and fivenines_server#608 can start NOW against this fixture** - server-first deploy order means nothing here blocks the backend build.

spuyet:
All three asks addressed in 22f56c4 (mqtt.py stays 100% covered; full suite 1519 passed).

**1. [P2] `disconnected_after_data` contract scenario — added.** New scenario in `mqtt_contract_payload.json`: connect -> two live messages -> disconnect -> snapshot. Pins the gated-staleness state exactly as spec'd: `status: "error"`, `error: "disconnected: Keep alive timeout"`, `connected_age_s: null`, per-monitor `subscribed_age_s: null`, and both topics **present** with growing last-observed ages (`last_live_seen_age_s: 35.0 / 30.0`) plus their captured payloads. The round-trip test replays it, and `test_contract_pins_retained_and_failure_semantics` now asserts the gate invariants so a regression fails loudly. fivenines_server#608 can build its e2e ingester replay against these bytes now.

**2. [P2] Absence semantics — pinned in the fixture `description`.** Documented the server-side rule that has no agent-visible payload: an absent `data["mqtt"]` means unconfigured **or** the collector raised (the per-collector wrapper at `collectors.py:146` swallows the exception and emits nothing), so the server disambiguates only from what it configured and must never synthesize a healthy-empty shape for a configured host. Same note for the per-broker variant: a broker whose construction fails is omitted from the `brokers` map (key present, that `broker_id` absent) -> treat configured-but-missing as collection-failed, never removed. The `null = crashed` wording is gone.

**3. [P3] `_wire_subscribe` now gates on the return code.** `subscribe()` returns `(MQTT_ERR_NO_CONN, None)` without raising, so arming was happening even when the SUBSCRIBE never went out. It now returns a bool; `_resubscribe_all` and `update_monitors` arm `subscribed_at` only for filters paho accepted (unchanged monitors keep their arm time). Two new tests cover the refusal path (`test_resubscribe_refused_does_not_arm`, `test_update_subscribe_refused_does_not_arm`). Added a TODOS bullet on why SUBACK/`on_subscribe` verification is deliberately skipped (Mosquitto ACKs ACL-denied subscriptions and silently drops deliveries, so the server's exact-topic staleness is the real safety net).

For the record, accepted as-is per the review: the dict-keyed `brokers`/`monitors`/`topics` map shape (agent repo is the contract source of truth) and the disabled-host-keeps-clients deferral (documented inline). Server-first deploy order unchanged.

spuyet:
Re-checked from the server side (fivenines_server#614 re-review): both asks verified landed in 22f56c4 - fixture scenario 7 (`disconnected_after_data`) is vendored byte-identically server-side, and the rc-aware subscribe arming is cleaner than the minimal ask (armed only for filters paho accepted). The server's ingest layer was reworked onto this contract and its specs now replay all 7 scenarios. Good to merge; release 1.12.0 after the server deploys.

PR #96, opened 2026-07-23
v1.13.0 feat: Ceph collector v2 (client I/O + recovery + per-pool + per-OSD)

### LINKED ISSUE #94: Ceph collection v2: client I/O + recovery rates, pool breakdown, per-OSD perf/fullness (contract 1.12.0)
_Ceph Datadog-parity plan - agent half. Server consumes this across 3 PRs: Five-Nines-io/Five-nines-server#615 (cluster scalars), Five-Nines-io/Five-nines-server#616 (per-pool), Five-Nines-io/Five-nines-server#617 (per-OSD)._

## Summary
Close the Ceph observability gap vs Datadog in one agent release: **client I/O throughput/IOPS**, **recovery/rebalance progress**, **nearfull/full OSD counts**, **per-pool usage**, and **per-OSD commit/apply latency + fullness**. Most of it is data we already fetch but curate out (`ceph status`, `ceph df`); only two new read-only commands are added (`ceph osd perf`, `ceph osd df`). Everything is presence-guarded server-side: the server deploys first, older agents keep working bit-identically.

## New payload fields (per entry in `clusters[]`)

**From `ceph status --format json` (already run - currently curated out):**
- `io`: `{read_bytes_sec, write_bytes_sec, read_op_per_sec, write_op_per_sec}` from pgmap. The mgr **omits** these keys when the cluster is idle - that is a true zero, so the agent normalizes: emit the object whenever status collection succeeded, absent keys filled with `0`.
- `recovery`: `{recovering_objects_per_sec, recovering_bytes_per_sec, misplaced_objects, misplaced_total, degraded_objects, degraded_total}` - same absent-means-0 normalization.
- `osd_fullness`: `{nearfull, full}` counts extracted from the `OSD_NEARFULL` / `OSD_FULL` health checks: use `summary.count` when present, else parse the leading integer of `summary.message` ("3 nearfull osd(s)"), else `null`. `null` = unknown - the server emits no sample; never fabricate a 0 from an unparseable check.

**From `ceph df --format json` (already run):**
- `pools`: `[{name, id, stored_bytes, objects, percent_used, max_avail_bytes}]`. `percent_used` forwarded **verbatim** as the 0-1 float ceph reports (the server converts for display). Cap at **200** entries + `pools_truncated` flag.

**New commands** (same target plumbing - keyring/conf/client-id/sudo - and the same per-command subprocess timeout as the existing three):
- `ceph osd perf --format json` → `osd_perf`: `[{id, commit_latency_ms, apply_latency_ms}]`. Handle both JSON shapes (Nautilus+ nests under `osdstats.osd_perf_infos`; older releases put `osd_perf_infos` top-level) and ship one normalized array.
- `ceph osd df --format json` → `osd_df`: `[{id, name, utilization, total_bytes, used_bytes, avail_bytes, status}]`. kb fields converted to bytes at the agent (x1024) so the contract stays bytes-native like `capacity`; `utilization` verbatim (0-100 float).
- Both arrays capped at **2048** entries + `osd_perf_truncated` / `osd_df_truncated` flags. The server skips per-OSD emission on a truncated tick (honesty over partial data), so the cap must be flagged, never silent.

## Collection flags
`collection` gains `perf_ok` and `osd_df_ok` (false on command failure / timeout / parse error). Each new command failing must NOT fail the core sections - independent isolation like the existing per-command handling. Existing semantics unchanged (`reachable` / `status_ok` / `df_ok` / `tree_ok`, fsid-null unreachable entries).

## Contract fixture (lockstep discipline)
- Update `tests/fixtures/ceph_contract_payload.json`: the reachable entry gains `io` / `recovery` / `osd_fullness` / `pools` / `osd_perf` / `osd_df` + the new flags; keep the fsid-null unreachable entry as-is. Extend `test_ceph.py::test_contract_fixture_round_trip`.
- The server vendors this fixture byte-identical (`spec/fixtures/ceph_contract_payload.json`). Change only in lockstep with the server PRs.

## Version / release
- Ships in the next minor (**1.13.0**). The server gates the extended surfaces on a distinct `FEATURES_SUPPORTED_VERSIONS["ceph_extended"] = "1.13.0"` key (the `redis_extended` precedent - upgrade hint, not a toggle gate); the base `ceph` capability stays 1.9.0.
- Release **after** the server-side tier-1 ingestion deploys (standard order; safe either way - the server presence-guards every new key).

## Out of scope
- No new config surface (rides the existing `config["ceph"]["clusters"]` targets).
- `ceph osd dump` ratios (nearfull/backfillfull/full) - the health-check counts cover alerting; revisit only if per-OSD threshold coloring needs the real ratios.

**Effort:** S-M (2 new commands + curation of 2 existing ones + fixture/tests).

spuyet:
## Summary

Closes #94. Second Datadog-parity tier for the Ceph collector, driven from the **same configured targets** (no new config surface). The agent half of the server's 3-PR ingestion (`#615` cluster scalars, `#616` per-pool, `#617` per-OSD). Every new field is additive and the server presence-guards each key, so older agents keep reporting bit-identically.

## New payload fields (per `clusters[]` entry)

**From `ceph status` (curated back in):**
- `io` `{read_bytes_sec, write_bytes_sec, read_op_per_sec, write_op_per_sec}` from pgmap; absent key -> `0` (mgr omits when idle = true zero), emitted whenever status succeeded.
- `recovery` `{recovering_objects_per_sec, recovering_bytes_per_sec, misplaced_objects, misplaced_total, degraded_objects, degraded_total}` -- same absent-means-0.
- `osd_fullness` `{nearfull, full}` from the `OSD_NEARFULL`/`OSD_FULL` health checks: `summary.count` -> leading int of `summary.message` -> `null`. **`null` = unknown; never a fabricated `0` from an unparseable check.**

**From `ceph df`:** `pools[]` `{name, id, stored_bytes, objects, percent_used, max_avail_bytes}`, `percent_used` verbatim (0-1 float), capped 200 + `pools_truncated`.

**Two new read-only commands** (same auth/timeout/cache plumbing):
- `ceph osd perf` -> `osd_perf[]` `{id, commit_latency_ms, apply_latency_ms}` -- handles **both** the Nautilus+ `osdstats.osd_perf_infos` nesting and the legacy top-level shape.
- `ceph osd df` -> `osd_df[]` `{id, name, utilization, total_bytes, used_bytes, avail_bytes, status}` -- kb x1024 to stay bytes-native, `utilization` verbatim (0-100).
- Both capped 2048 + `osd_perf_truncated`/`osd_df_truncated` (server skips per-OSD emission on a truncated tick).

`collection` gains `perf_ok` + `osd_df_ok`. Each new command is **independently isolated**: a failure/timeout logs and leaves its flag `False` + value `None` without touching the status-derived core sections.

## Version (please confirm server-side)

The issue text says ship at **1.12.0**, but that slot was taken by the MQTT collector (#93). This ships at the next free minor, **1.13.0**. **The server's `FEATURES_SUPPORTED_VERSIONS["ceph_extended"] must be `"1.13.0"`, not `1.12.0`.** Base `ceph` capability stays 1.9.0.

## Design note: `osd_fullness` absent -> null

An absent `OSD_NEARFULL`/`OSD_FULL` check yields `null` (per the issue's "else null / never fabricate a 0 from an unparseable check"), so a healthy cluster reports `{nearfull: null, full: null}` and the server emits no sample. If a continuous `0` line during health is preferred, that's a one-line change in `_extract_fullness_count`.

## Testing

- `tests/test_ceph.py`: round-trip contract test extended + 33 new unit tests covering every new branch. **82 passing, `ceph.py` at 100% coverage.**
- Full suite (libvirt mocked, deps installed locally): **1552 passed, 2 skipped, 0 failures** -- confirms the collector registry that imports `ceph_metrics` is unaffected.
- Adversarial review: **SHIP** -- no substantiated correctness/contract defects across contract-shape, dual JSON shapes, truncation boundaries, absent/null/zero semantics, defensive parsing, isolation, and bool-as-int handling.
- Clean pyflakes; ASCII-only.

## Lockstep / pending (server-first deploy)

The shared `tests/fixtures/ceph_contract_payload.json` is the source of truth; the server vendors it byte-identical at `spec/fixtures/ceph_contract_payload.json`. Pending server work: ingesters #615/#616/#617, `ceph_extended = "1.13.0"`, fixture copy, `api_collect_ceph_spec.rb`. Safe to deploy either order (every key presence-guarded), but the standard order is server-first.

**Caveat (carried from v1):** parsing was built from documented `ceph -f json` shapes and fixtures, **not validated against a live cluster** -- worth a real-cluster field check around deploy.

### DIFF
--- a/fivenines_agent/ceph.py
+++ b/fivenines_agent/ceph.py
@@ -33,6 +33,7 @@ import json
 import re
 import shutil
 import subprocess
+import time
 
 from fivenines_agent.cache import TTLCache
 from fivenines_agent.debug import log
@@ -42,6 +43,16 @@ from fivenines_agent.subprocess_utils import get_clean_env
 CONNECT_TIMEOUT = 5  # seconds, --connect-timeout passed to the ceph CLI
 SUBPROCESS_TIMEOUT = 15  # seconds, hard subprocess kill (a wedged mon can hang)
 CACHE_TTL = 30  # seconds, per (cluster, command)
+
+# Per-cluster wall-clock budget. Collectors run serially in the main loop and
+# ceph failures are not cached (a persistently-down cluster is retried every
+# tick), so a wedged mon would otherwise block the tick for up to 5 commands x
+# SUBPROCESS_TIMEOUT per cluster. Once this many seconds have elapsed for a
+# cluster, the enrichment tail (df/tree/perf/osd df) is shed with its *_ok flag
+# left False -- capping the worst case near the v1 3-command budget so a
+# multi-cluster host cannot blow past the server's offline window. status is
+# core (drives reachability) and always runs regardless.
+ENRICHMENT_DEADLINE = 30  # seconds
 
 # Per-cluster emission caps. pools is small in practice; the per-OSD arrays can
 # be large on a big cluster, so they are capped harder. A hit sets the matching
@@ -102,13 +113,17 @@ def _poll_cluster(cluster):
     """Poll one cluster and build its contract payload.
 
     status drives reachability + most metrics (health, mon, osd, pg, io,
-    recovery, osd_fullness). df, osd tree, osd perf and osd df are each partial
-    and independently isolated: a failure leaves that command's *_ok flag False
-    and its value None without making the whole cluster unreachable.
+    recovery, osd_fullness) and always runs. df, osd tree, osd perf and osd df
+    are the enrichment tail: each is partial and independently isolated (a
+    failure leaves that command's *_ok flag False and its value None without
+    making the whole cluster unreachable), and each is skipped once the
+    per-cluster ENRICHMENT_DEADLINE passes so a wedged mon cannot block the
+    serial main loop for the full 5 x SUBPROCESS_TIMEOUT.
     """
     name = cluster.get("name") or "ceph"
     base = _base_args(cluster)
     result = _empty_result(name)
+    deadline = time.monotonic() + ENRICHMENT_DEADLINE
 
     status, error = _run_ceph_cached(base, ["status"], name)
     if error:
@@ -125,54 +140,76 @@ def _poll_cluster(cluster):
     result["collection"]["status_ok"] = True
     _apply_status(result, status)
 
-    df, error = _run_ceph_cached(base, ["df"], name)
-    if error:
-        log(
-            "ceph df failed for cluster {}: {}".format(name, error.get("message")),
-            "error",
-        )
-    else:
-        result["collection"]["df_ok"] = True
-        result["capacity"] = _parse_capacity(df)
-        result["pools"], result["pools_truncated"] = _parse_pools(df)
+    if not _past_deadline(deadline, name):
+        df, error = _run_ceph_cached(base, ["df"], name)
+        if error:
+            log(
+                "ceph df failed for cluster {}: {}".format(name, error.get("message")),
+                "error",
+            )
+        else:
+            result["collection"]["df_ok"] = True
+            result["capacity"] = _parse_capacity(df)
+            result["pools"], result["pools_truncated"] = _parse_pools(df)
 
-    tree, error = _run_ceph_cached(base, ["osd", "tree"], name)
-    if error:
-        log(
-            "ceph osd tree failed for cluster {}: {}".format(
-                name, error.get("message")
-            ),
-            "error",
-        )
-    else:
-        result["collection"]["tree_ok"] = True
-        result["hosts"] = _parse_host_osd_counts(tree)
+    if not _past_deadline(deadline, name):
+        tree, error = _run_ceph_cached(base, ["osd", "tree"], name)
+        if error:
+            log(
+                "ceph osd tree failed for cluster {}: {}".format(
+                    name, error.get("message")
+                ),
+                "error",
+            )
+        else:
+            result["collection"]["tree_ok"] = True
+            result["hosts"] = _parse_host_osd_counts(tree)
 
-    perf, error = _run_ceph_cached(base, ["osd", "perf"], name)
-    if error:
-        log(
-            "ceph osd perf failed for cluster {}: {}".format(
-                name, error.get("message")
-            ),
-            "error",
-        )
-    else:
-        result["collection"]["perf_ok"] = True
-        result["osd_perf"], result["osd_perf_truncated"] = _parse_osd_perf(perf)
+    if not _past_deadline(deadline, name):
+        perf, error = _run_ceph_cached(base, ["osd", "perf"], name)
+        if error:
+            log(
+                "ceph osd perf failed for cluster {}: {}".format(
+                    name, error.get("message")
+                ),
+                "error",
+            )
+        else:
+            result["collection"]["perf_ok"] = True
+            result["osd_perf"], result["osd_perf_truncated"] = _parse_osd_perf(perf)
 
-    osd_df, error = _run_ceph_cached(base, ["osd", "df"], name)
-    if error:
-        log(
-            "ceph osd df failed for cluster {}: {}".format(
-                name, error.get("message")
-            ),
-            "error",
-        )
-    else:
-        result["collection"]["osd_df_ok"] = True
-        result["osd_df"], result["osd_df_truncated"] = _parse_osd_df(osd_df)
+    if not _past_deadline(deadline, name):
+        osd_df, error = _run_ceph_cached(base, ["osd", "df"], name)
+        if error:
+            log(
+                "ceph osd df failed for cluster {}: {}".format(
+                    name, error.get("message")
+                ),
+                "error",
+            )
+        else:
+            result["collection"]["osd_df_ok"] = True
+            result["osd_df"], result["osd_df_truncated"] = _parse_osd_df(osd_df)
 
     return result
+
+
+def _past_deadline(deadline, name):
+    """True once this cluster's enrichment budget is spent (logged once per hit).
+
+    Checked before each enrichment command; the commands already run core-first
+    (df -> tree -> perf -> osd df) so what gets shed is always the least-critical
+    tail. The shed commands keep their *_ok False, which the server treats as
+    "section not collected" -- identical to a per-command failure.
+    """
+    if time.monotonic() < deadline:
+        return False
+    log(
+        "ceph: enrichment deadline reached for cluster {}, "
+        "skipping remaining commands".format(name),
+        "error",
+    )
+    return True
 
 
 def _empty_result(name):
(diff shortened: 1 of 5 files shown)


fivenines_agent/ceph.py
```diff
@@ -354,6 +426,103 @@ def _parse_pg(status):
...
+            pgmap.get("recovering_bytes_per_sec")
+        ),
+        "misplaced_objects": _num_or_zero(pgmap.get("misplaced_objects")),
+        "misplaced_total": _num_or_zero(pgmap.get("misplaced_total")),
+        "degraded_objects": _num_or_zero(pgmap.get("degraded_objects")),
+        "degraded_total": _num_or_zero(pgmap.get("degraded_total")),
+    }
+
+
+def _parse_osd_fullness(status):
```
spuyet, line 518: **Contract semantics: an absent check should be `0`, not `null` - the current mapping inverts Ceph's own semantics and contradicts this PR's io/recovery rule.**

Ceph health checks are present **iff the condition is active**: a valid `checks` dict without `OSD_NEARFULL` is the mon positively asserting "0 nearfull OSDs" - the exact analogue of pgmap omitting `read_bytes_sec` on an idle cluster, which `_parse_io` (correctly) normalizes to 0 a few functions up. With the current mapping a healthy cluster reports `{nearfull: null, full: null}` forever, so the server emits no sample and the `ceph_osd_nearfull` series only exists while the cluster is in trouble:

- charts get no flat 0 line - "healthy" and "no data" become indistinguishable;
- any future threshold alert on the count could fire but never resolve on a 0 - recovery becomes a series *disappearance*, which is the exact ambiguity the null was meant to avoid;
- the docstring's rationale ("a wrong 0 would silence a real nearfull condition") is right for a **present-but-unparseable** check, but does not apply to an absent one.

Suggested split - absent key in a valid checks dict -> `0` (true zero); present but unparseable -> `None` (unknown); health/checks malformed -> `None` for both:

```python
def _parse_osd_fullness(status):
    health = status.get("health")
    checks = health.get("checks") if isinstance(health, dict) else None
    if not isinstance(checks, dict):
        return {"nearfull": None, "full": None}
    return {
        "nearfull": (_extract_fullness_count(checks["OSD_NEARFULL"])
                     if "OSD_NEARFULL" in checks else 0),
        "full": (_extract_fullness_count(checks["OSD_FULL"])
                 if "OSD_FULL" in checks else 0),
    }
```

Fixture impact: the reachable entry becomes `"full": 0` (checks dict valid, no OSD_FULL check) - the null case moves to a unit test (present-unparseable), and #615's "null = no sample" keeps covering exactly that case. The server ingester is not built yet, so this is the free moment to fix the contract; once #615 ships, changing it means a lockstep dance.

fivenines_agent/ceph.py
```diff
@@ -123,10 +148,40 @@ def _poll_cluster(cluster):
         result["collection"]["tree_ok"] = True
         result["hosts"] = _parse_host_osd_counts(tree)
 
+    perf, error = _run_ceph_cached(base, ["osd", "perf"], name)
```
spuyet: **Serial tick budget: worst case is now 75s per cluster, every tick.**

`collect_metrics` runs collectors serially in the main loop, ceph failures are (deliberately, and correctly) not cached, and each command has a 15s kill. A wedged mon therefore blocks the whole tick 5 x 15s = 75s per cluster (was 3 x 15s = 45s in v1), on **every** tick, multiplied by the number of configured clusters. Concrete stake: at interval 60s with the server's offline threshold at ~2 x interval, a host with two wedged clusters (150s of blocking) can miss the /collect window and get false-flagged offline - the monitored cluster's outage taking down the *host's* monitoring.

The command order already puts core first (status -> df -> tree -> perf -> osd df), so a cheap mitigation composes well: a per-cluster deadline in `_poll_cluster` (e.g. stop issuing the remaining commands once ~30s have elapsed for this cluster, leaving their `*_ok` flags False - a state the server already treats as "section not collected"). Alternatively a shorter `SUBPROCESS_TIMEOUT` for the two new enrichment commands.

Not a blocker - the pattern predates this PR - but the PR widens the worst case by 66% and multi-cluster hosts pay it multiplied, so worth either fixing here or filing.

spuyet:
## Review (server-side counterpart perspective)

Read the full diff plus the collector dispatch path (`collect_metrics` is serial, `_run_ceph_cached` caches successes only). The contract matches #94 on everything I checked: field names/shapes, caps + honest truncation flags set from raw lengths, per-command isolation with `*_ok` semantics, verbatim `percent_used` (0-1) and `utilization` (0-100), kb -> bytes conversion, both `osd perf` JSON shapes, additive-null unreachable entry, bool-as-int rejection where the agent computes. Tests are thorough (caps via monkeypatch, isolation, both shapes, fixture round-trip with correct token routing).

Two findings posted inline:

1. **Contract fix before the server builds on it** (`_parse_osd_fullness`): an absent `OSD_NEARFULL`/`OSD_FULL` check in a valid checks dict is Ceph asserting a **true zero**, not "unknown" - same reasoning as the pgmap absent-means-0 rule this PR applies to io/recovery. As shipped, the nearfull/full series would only exist during incidents (no flat 0 line; a future count alert could fire but never resolve). Suggested absent -> 0 / present-but-unparseable -> null split inline; fixture reachable entry becomes `"full": 0`. This is the "one-line change" the PR body itself offers - I am voting for it, now, while the server ingester does not exist yet.
2. **Tick budget** (`_poll_cluster`): worst case is now 5 x 15s = 75s of serial main-loop blocking per cluster per tick when a mon wedges (failures uncached by design), which can push a multi-cluster host past the server's offline window. Suggested per-cluster deadline shedding the enrichment tail (the `*_ok` flags already degrade cleanly). Fix here or file it.

Nit (take or leave): `_parse_osd_df` could skip nodes whose `type` is present and != `"osd"` - free future-proofing if the command ever moves to `osd df tree`, where host/root buckets would ingest as phantom OSDs.

**Version shift acknowledged: 1.13.0 confirmed server-side.** I have updated the three server issues (Five-Nines-io/Five-nines-server#615/#616/#617) and #94 so `FEATURES_SUPPORTED_VERSIONS["ceph_extended"]` gets pinned to `"1.13.0"` - no server work had started, so nothing else to reconcile. Fixture stays byte-identical lockstep once finding 1 is settled.

fivenines_agent/ceph.py
spuyet: Fixed in edaf625. Adopted the split exactly: absent check in a valid `checks` dict -> `0` (true zero), present-but-unparseable -> `None`, malformed health/checks -> `None` for both. `_extract_fullness_count` now only ever sees present checks (docstring updated). Fixture reachable entry is now `"full": 0`; the null case moved to a unit test (`test_parse_osd_fullness_present_unparseable_is_null`), and `test_parse_osd_fullness_absent_checks_are_zero` pins the true-zero path. ceph.py still 100% covered.

fivenines_agent/ceph.py
spuyet: Fixed in edaf625 with the per-cluster-deadline option. New `ENRICHMENT_DEADLINE = 30`: `_poll_cluster` records `time.monotonic() + 30` and checks `_past_deadline` before each enrichment command (df/tree/perf/osd df), shedding the tail with `*_ok` left False once the budget is spent. status stays core and always runs. Worst case is bounded near the v1 3-command budget (~deadline + one in-flight SUBPROCESS_TIMEOUT) instead of 5x15s, and healthy clusters (cache hits/fast commands) never trip it. Covered by `test_poll_cluster_deadline_sheds_enrichment_tail`.

spuyet:
All three review points addressed in edaf625 (replies inline on the two findings):

1. **osd_fullness absent -> 0** (contract fix): absent check in a valid `checks` dict = true zero; present-but-unparseable = `null`; malformed health/checks = `null`. Fixture reachable entry is now `full: 0`.
2. **Per-cluster deadline**: `ENRICHMENT_DEADLINE = 30s` sheds the df/tree/perf/osd-df tail (`*_ok` False) so a wedged mon can't blow the tick past the offline window.
3. **Nit taken**: `_parse_osd_df` skips nodes whose `type` is present and `!= "osd"` (phantom CRUSH buckets if ever pointed at `osd df tree`); typeless plain-command rows kept. Round-trip CLI nodes now carry `type: "osd"` for realism.

`ceph.py` stays at 100% coverage (+3 tests, 85 total); full suite 1555 passed. Fixture is settled, so the byte-identical server copy can be vendored whenever #615/#616/#617 land. Thanks for the server-side pass and the 1.13.0 confirmation.

fivenines_agent/ceph.py
spuyet: Re-checked on edaf6250: exactly the suggested split, and the docstrings now carry the rationale on both sides (absent = the mon asserting a true zero; None reserved for present-but-unparseable and malformed health/checks). Fixture reachable entry is "full": 0 and the present-unparseable case moved to a dedicated unit test as discussed. Independently ran tests/test_ceph.py on the commit: 85 passed. Resolved.

fivenines_agent/ceph.py
spuyet: Re-checked on edaf6250: deadline implemented as discussed - status (core) always runs and its 15s timeout can never trip the 30s gate before df, so the v1 core sections are never shed on a healthy-but-slow cluster; worst case is now ~30s elapsed + one in-flight timeout = ~45s/cluster, back to the v1 budget. The shed path is proven by test (only status issued, all four tail flags False). Resolved.

spuyet:
## Re-review: LGTM

edaf6250 addresses all three points from the first pass, verified against the actual diff plus an independent local run (no CI is wired on this repo):

1. **osd_fullness semantics** - exact absent->0 / present-unparseable->None / malformed->None split, fixture at `"full": 0`, dedicated unit test for the None case. Resolved.
2. **Tick budget** - `ENRICHMENT_DEADLINE = 30` gates the four tail commands, status always runs, worst case back to ~45s/cluster (the v1 budget); shed path proven by test. Resolved.
3. **Nit taken** - `_parse_osd_df` skips typed non-osd nodes, keeps typeless ones, tested with a root bucket.

Independent verification: cloned the branch, `pytest tests/test_ceph.py` -> **85 passed** (includes the contract round-trip); `ceph.py` compiles clean.

Good to merge from my side. Lockstep reminder for the server work (#615-#617): vendor the fixture from **this commit or later** (the `"full": 0` entry), and `ceph_extended = "1.13.0"` (issues already updated). Standard order: server ingestion deploys first, then release 1.13.0.

PR #102, opened 2026-07-30
v1.14.0 feat: Docker image OS-package inventory via the archive API (#101)

### LINKED ISSUE #101: Docker image package inventory: extract OS package lists from images via the archive API (image vulnerability scanning, phase 1)
Server-side counterpart of Docker **image** vulnerability scanning (phase 1). The server already syncs the OSV database for Debian / Ubuntu / Alpine / AlmaLinux / Red Hat and already has the matcher (`Osv::PackageScanner`) and the distro mapper (`Osv::EcosystemMapper`, which consumes exactly `"debian:13"` - the format `packages.py:get_distro()` already produces). The only missing input is **the package list inside a container image**, which only the agent can produce.

Requested by several paying customers.

## Mechanism: the Docker archive API, not `docker exec`

`GET /containers/{id}/archive?path=<p>` asks the daemon to tar a path from the container's filesystem and stream it back. It is what `docker cp` uses. The daemon mounts the container's layers on demand, so:

- **no binary is needed inside the image** (`dpkg-query`, `rpm`, `sh` - none)
- **works on a stopped, and even never-started, container**
- **driver-agnostic**
- **a missing path returns a clean 404**, which *is* the distroless/scratch signal
- **no process runs in the customer's container** (unlike `docker exec`)

Verified on Docker 25.0.3 / overlay2 against a `created` (never started) `postgres:16` container:

```
$ curl --head --unix-socket /var/run/docker.sock \
    "http://localhost/v1.44/containers/$CID/archive?path=%2Fetc%2Fos-release"
HTTP/1.1 200 OK
X-Docker-Container-Path-Stat: eyJuYW1lIjoib3MtcmVsZWFzZSIsInNpemUiOjIx...
```
```json
{"name":"os-release","size":21,"mode":134218239,"linkTarget":"/usr/lib/os-release"}
```

**Measured cost** (`postgres:16`, Debian 13 base): `/var/lib/dpkg/status` = 137 263 bytes, tar 139 264, **gzipped 37 460**, `get_archive` **0.06 s**, **142 installed packages**. `postgres:16-alpine`: `/lib/apk/db/installed`, **45 packages, 27 625 bytes gzipped**. And since an image digest is immutable, this is **once per image, forever** - not per tick like host packages.

### Gotcha found while verifying: `/etc/os-release` is a symlink and the GET does NOT follow it

A direct `GET ?path=/etc/os-release` returns a tar containing a **symlink entry**, not the file:

```
nom='os-release' type=SYMLINK taille=0 linkname='../usr/lib/os-release'
# python tarfile: KeyError: "linkname '../usr/lib/os-release' not found"
```

So the target must be re-requested. Use the **`linkTarget` from `X-Docker-Container-Path-Stat`** (absolute: `/usr/lib/os-release`), *not* the tar entry's `linkname` (relative: `../usr/lib/os-release`). `container.get_archive(path)` returns `(bits, stat)`, so the stat is available from the same GET - re-request once when `stat["linkTarget"]` is set. An agent that skips this concludes every Debian image has no `os-release` and reports the whole fleet unscannable.

---

## HAZARD 1: do not put the new config key inside `config["docker"]`

`collectors.py:182-202` splats `config["docker"]` as `**kwargs` into `docker_metrics`, and `docker_metrics(socket_url=None)` (`docker.py:361`) has **no `**kwargs`**. So adding `inventory_requests` to that dict makes **every existing agent** raise `TypeError: docker_metrics() got an unexpected keyword argument`, swallowed by `_collect_with_telemetry` (`collectors.py:152-180`) into `data["docker"] = None`.

Combined with the server's never-prune-on-null rule, that means: **Docker metrics go dark and container-state rows freeze forever on every agent below the new version.** Silent, fleet-wide, and it would look like a server bug.

Two consequences for this issue:

1. **The server must use a TOP-LEVEL config key**, read directly from `config` and never splatted - the `capture_logs` pattern. Not nested under `docker`. Zero version coupling, no breaking window.
2. **Add `**_kwargs` to `docker_metrics` anyway**, matching the existing precedent (`systemd.py:1298-1304` *"Extra kwargs are accepted but ignored so new server-side config keys do not break older agents"*, `ceph.py:72` `def ceph_metrics(clusters=None, **_)`). `docker.py` is the odd one out and this is exactly the trap it sets.

---

## HAZARD 2: rootless Docker is NOT supported today, and this feature is where it hurts most

Two different things get called "non-root Docker" and only one of them works:

| Setup | Status |
|---|---|
| **Agent as non-root (`User=fivenines`) + root daemon**, `fivenines` in the `docker` group, socket `/var/run/docker.sock` | **Works today** (`README.md:168-176`, `usermod -aG docker fivenines`). This feature inherits it for free - same socket, same group, **no new privilege, no new group, no sudoers entry**. |
| **Rootless Docker** (daemon runs as an unprivileged user, socket at `$XDG_RUNTIME_DIR/docker.sock`, e.g. `/run/user/1000/docker.sock`) | **Broken today**, for the whole Docker collector - not just this feature. |

### Why rootless is broken

`permissions.py:623-646` hardcodes the path:

```python
def _can_access_docker(self):
    docker_socket = "/var/run/docker.sock"
    if not os.path.exists(docker_socket):
        self._set_reason(f"{docker_socket} does not exist")
        return False
```

It ignores `config["docker"]["socket_url"]` **and** `DOCKER_HOST`. And a False capability is a hard gate: `_is_capability_gated` (`collectors.py:131-150`) skips the collector entirely, so `data["docker"]` is **absent** - not even `null`. Net effect on a rootless host: the operator sets `docker_socket_url` in the UI, the agent could reach it, and nothing is ever collected. The banner says `[-] Docker (requires docker group)`, which is a misleading reason.

`_can_access_libvirt` was already rewritten to mirror the collector's real connect instead of guessing a path (see the comment at `permissions.py:683-690`) - that is the precedent to copy.

### Why it matters more for image inventory than for metrics

Under rootless (and under `userns-remap`), the container's files on disk live in `~/.local/share/docker` and are owned by **subordinate UIDs from `/etc/subuid`** (100000+). A process outside that user namespace **cannot read them at all** - so the "read `/var/lib/docker/overlay2/<id>/merged/...` from the host" approach is not merely ugly there, it is impossible. The daemon runs *inside* the namespace where it is root, so the archive API works unchanged. **Rootless is the case that makes the archive API the only viable design**, which is a good reason to make it work rather than document it as unsupported.

### Scope for this issue (be honest about the limit)

- **Fix the probe**: `_can_access_docker` resolves the socket the same way the collector does - configured `socket_url` → `DOCKER_HOST` → `/var/run/docker.sock` → `$XDG_RUNTIME_DIR/docker.sock` - and reports the real reason (`not accessible (readable=..., writable=...)` naming the path it actually tried). This is the actual bug: an operator who *can* grant access still cannot use it.
- **Document the rootless setup**: a systemd drop-in with `Environment=DOCKER_HOST=unix:///run/user/<uid>/docker.sock`, or run the agent as the rootless user. The unit ships `User=fivenines` with no `XDG_RUNTIME_DIR` and no `DOCKER_HOST` passthrough.
- **Do NOT promise auto-discovery of another user's rootless socket.** `/run/user/<uid>` is 0700-owned by that uid; `fivenines` cannot reach it, and that is correct. Say so.
- **Note the user-install trap**: `fivenines_setup_user.sh` runs the agent from `@reboot` cron, where `XDG_RUNTIME_DIR` is typically unset - so `docker.from_env()` works in an interactive shell and fails under cron. Classic rootless false negative.

---

## What to extract (phase 1: Debian / Ubuntu / Alpine)

Per requested image digest, in order:

1. `/etc/os-release` (following `linkTarget`) → `ID` + `VERSION_ID` → `"debian:13"`. **Reuse `packages.py:get_distro()`'s parsing** - same keys, same lowercase/strip-quotes rules, same `"unknown"` fallback. Do not write a second parser.
2. Branch on the distro family:
   - dpkg (`debian`, `ubuntu`): `/var/lib/dpkg/status` - RFC822 stanzas split on blank lines; keep `Package:` / `Version:` **and require `Status:` to contain `install ok installed`** (otherwise a removed-but-config-files-remain entry is reported as installed). Skip continuation lines (leading space/tab).
   - apk (`alpine`): `/lib/apk/db/installed` - same stanza shape, keys `P:` (package) and `V:` (version).
3. `packages_hash` via the existing `get_packages_hash()` (`packages.py:262-265`) so server-side dedupe is byte-compatible with the host path.

**RPM is out of scope** (phase 1.5, separate issue): `rpmdb.sqlite` retrieves identically, but the rows hold binary RPM header blobs that need decoding (tags 1000/1001/1002/1003), and RHEL 8-era images use Berkeley DB. Report `unsupported` with a reason for now - never silently empty.

## Where it goes in the code

- `docker.py:87-105` `_image_tags_and_digests` already calls `client.images.get(image_id)` **once per distinct image id per tick**, memoized in the per-tick cache created at `docker.py:279`. That cache is the natural hook: it already enumerates the distinct digests on the host and maps each to a container.
- **New module** (`docker_image_inventory.py`) for the extraction + parsers, so `docker.py` stays a metrics/state collector.
- **Run it out-of-band, not on the collect loop.** `/packages` and `/systemd_inventory` are posted synchronously (`agent.py:256-257`) and `wd.notify()` exists at `agent.py:249-255` precisely because that can stretch a tick toward `WatchdogSec=90`. N archive fetches + tar parsing + a POST belongs in a bounded worker: copy `LogUploader` + `CaptureCoordinator` (`log_uploader.py`, `log_capture.py`) - bounded `SynchronizationQueue`, per-job isolation, queue-full shedding, in-flight/`already-done` tracking so a re-requested digest is not re-extracted.

## POST contract

New `Synchronizer.send_image_packages` following `_post` verbatim (gzip, `Bearer`, 3 retries with linear backoff, 200 = success):

```json
{"image_id": "sha256:...",
 "distro": "debian:13",
 "packages_hash": "<sha256, same formula as get_packages_hash>",
 "packages": [{"name": "openssl", "version": "3.0.11-1~deb12u2", "ecosystem": null}],
 "errors": [{"step": "os_release", "type": "not_found", "message": "..."}]}
```

`ecosystem: null` means "the image's OS ecosystem". It is there so a future phase can send `{"name": "lodash", "version": "4.17.20", "ecosystem": "npm"}` on the same endpoint without a contract break. Two extra characters now, no `/image_packages_v2` later.

**The server answers 200 for every definitive outcome, including refusals** (`{"status":"ok","accepted":false,"reason":"unsupported_distro"}`) so the agent marks the digest done and stops retrying. Non-200 is transient only. Same posture as `/logs`.

## Caps and the honesty contract

- per-file cap **5 MB**, enforced **while streaming** (running `buf.tell()` check, abort mid-stream) - not after
- **≤ 2000 packages** per image, **≤ 3 images per tick**
- scrub NUL / invalid UTF-8 from names and versions before sending
- **an extraction failure is never an empty package list.** No `os-release` → `unsupported`. No package DB → `unsupported`. 404 / container removed mid-extraction / cap exceeded / parse failure → `unscannable` with a structured `errors[]` entry. The server renders these as "not scannable, \<reason\>"; an empty `packages` array with an empty `errors` array would render as **"0 vulnerabilities"**, which is a false all-clear on a security feature.

## Tests

- New scenarios in `tests/fixtures/docker_image_inventory_contract_payload.json`, same shape as `docker_contract_payload.json` (`description` naming the source of truth, `agent_min_version`, `error_path_audit`, `scenarios`). **Bump `FEATURES_SUPPORTED_VERSIONS["docker_image_inventory"]` on the server to match exactly** - the note at the top of the docker fixture already states that rule.
- Scenarios to pin: `debian_slim`, `alpine`, `os_release_symlink` (the gotcha above), `distroless_no_package_db`, `container_removed_mid_extraction` (404), `file_over_cap`, `rpm_unsupported`, `dpkg_status_with_deinstalled_entry`, `malformed_stanza`.
- Mock at the SDK boundary as `tests/test_docker.py:825-852` does, with `get_archive` returning `(iter([tar_bytes]), stat_dict)`.
- 100% coverage per `CLAUDE.md:144`.
- **Add a `docker` key to `_DRY_RUN_CONFIG`** (`agent.py:72-100`) - it has none today, so `--dry-run` never exercises the Docker collector at all.

## Out of scope

- RPM (phase 1.5) and language ecosystems / lockfiles (phase 2, where the archive API stops helping because globbing needs a path known in advance)
- image size / creation date / layer list (one field away from `client.images.get(image_id).attrs`, but nothing consumes them yet)
- images with no container at all (would need `docker create`; we only ever learn digests from containers, so it cannot happen)
- the Docker events API for short-lived containers (already tracked, `TODOS.md:171-186`)

## Acceptance criteria

- [ ] `docker_metrics` accepts and ignores unknown kwargs (`**_kwargs`), matching `systemd.py` / `ceph.py`
- [ ] `_can_access_docker` resolves the socket like the collector does (`socket_url` → `DOCKER_HOST` → `/var/run/docker.sock` → `$XDG_RUNTIME_DIR/docker.sock`) and reports the path it actually tried
- [ ] Container-state + metrics collection verified working on a **rootless** daemon (this is currently broken and is the regression test for it)
- [ ] Verified on a `created`/never-started container and on a stopped container
- [ ] `/etc/os-release` symlink resolved via `linkTarget`; a Debian image reports a distro, not `unsupported`
- [ ] dpkg `Status: install ok installed` filter asserted against a fixture containing a deinstalled entry
- [ ] Extraction runs off the collect loop; a 3-image tick does not extend the tick measurably
- [ ] Every failure path produces a structured `errors[]` entry and never an empty-and-clean payload
- [ ] README gains a rootless Docker section (drop-in `DOCKER_HOST`, or run as the rootless user) and states the `/run/user/<uid>` limitation plainly

spuyet:
Agent half of Docker **image** vulnerability scanning (phase 1). The server already syncs OSV and has both the matcher (`Osv::PackageScanner`) and the distro mapper (`Osv::EcosystemMapper`, which consumes exactly the `"debian:13"` format `packages.py` already produces). The only missing input is the package list *inside* an image, which only the agent can produce.

Closes #101

## Mechanism

`GET /containers/{id}/archive` (`container.get_archive`), not `docker exec`. The daemon mounts the container's layers on demand, so no binary is needed inside the image, it works on stopped and never-started containers, it is driver-agnostic, a missing path is a clean 404, and nothing runs in the customer's container. Under rootless/`userns-remap` the layer files on disk belong to subordinate UIDs a host process cannot read at all -- the daemon runs *inside* that namespace, so this is the only design that works there.

Because an image digest is immutable, extraction is **once per image forever**, not per tick.

## Verified live

Against a real daemon, matching the issue's own measurements exactly:

| image | result |
|---|---|
| `postgres:16` | `debian:13`, **142 packages** |
| `postgres:16-alpine` | `alpine:3.24.1`, **45 packages** |
| exited container | extracted fine (stopped-container support) |
| `node-exporter` (scratch) | `distro: null` + `os_release/not_found` -- never empty-and-clean |

All 7 images in 0.16s. The `/etc/os-release` symlink resolved correctly (a Debian image reports a distro, not `unsupported`), and the honesty contract held on every payload.

## The honesty contract

A false all-clear on a security feature is worse than no feature. An extraction failure is **never** an empty package list: every failure path records a structured `errors[]` entry, including the subtle case where the DB reads fine but parses to nothing. `packages: []` with `errors: []` is the one shape this module must never produce -- the server renders it as "0 vulnerabilities".

## Hostile-image hardening

A customer may run a compromised image, so the extractor treats image content as attacker-controlled. Each of these was **measured** before and after:

| issue | before | after |
|---|---|---|
| package cap applied after materializing every stanza | 5 MB apk DB of 9-byte stanzas -> ~580k dicts, **142.8 MiB peak** | **0.4 MiB** (capped during parsing, lazy stanza iterator) |
| `tarfile.open` default mode = transparent decompression | 45,944-byte xz tar passed the byte cap, `read()` returned **314,583,040 bytes** | refused (`mode="r:"` + bounded read) |
| streaming cap checked after the write | up to +2 MiB overshoot (docker-py streams 2 MiB chunks) | checked before the write; buffer never exceeds the cap |
| `_scrub` removed only NUL | image-controlled newlines/ANSI escapes reached the payload and logs | C0/C1 stripped, length-bounded |
| truthiness checked on the raw name | `Package: \x00` shipped `{'name': ''}` **with `errors: []`** -- suppressing the guard above | entry dropped, guard stays armed |

## Both hazards from the issue

**HAZARD 1** -- the enabling key is **top-level `image_inventory`**, never nested under `docker` (`collectors.py` splats that dict as `**kwargs`, so a nested key would `TypeError` on every older agent, be swallowed into `data["docker"] = None`, and freeze container-state rows fleet-wide). `docker_metrics` also gained `**_kwargs` as a belt-and-braces guard, matching `systemd.py`/`ceph.py`.

**HAZARD 2** -- `_can_access_docker` no longer hardcodes `/var/run/docker.sock`. Both the probe **and the collector** now resolve `socket_url -> DOCKER_HOST -> /var/run/docker.sock -> $XDG_RUNTIME_DIR/docker.sock`. That last step had to be added to the collector too: `docker.from_env()` has no XDG fallback, so a probe that checked XDG while the client did not would just move the bug -- reporting the capability available and then failing to connect. The probe reports the path it actually tried.

## Cross-repo contract (server-first)

`tests/fixtures/docker_image_inventory_contract_payload.json` is the source of truth. **Vendor it byte-identical -- do not hand-author the server copy.** That inversion is what shipped 6 dead gauges on Ceph #615 and was caught again on MQTT #608.

It now documents three things a server dev cannot infer from the payloads alone:

- **`config`** -- the enabling key, that it is top-level, that `docker` must also be on, and the `image_id` join key back to `data["docker"]["containers"]`.
- **`response_contract`** -- the agent's success test is a **truthy JSON body**, not a 200. A `head :ok` (empty body) or `{}` is treated as transient and re-extracts *forever*. Refusals must be 200 with a non-empty body; 4xx must not be used for permanent refusals; the payload must be durably persisted before responding 200, because `mark_done` is one-shot.
- **`errors_contract`** -- the CLOSED `type`/`step` enums. Three of six emitted types appear in no scenario, and **`package_cap` is the high-stakes one**: it means `packages` is a truncated prefix, so a server with no case for it renders a partial scan as complete. A drift test now fails if the code emits a type the contract does not declare.

Server side still needs: the `/image_packages` ingester, `FEATURES_SUPPORTED_VERSIONS["docker_image_inventory"] = "1.14.0"`, and the byte-identical fixture copy. Deploy order is safe either way (agent-first is inert because the config key is never set; server-first is inert because old agents never POST).

## Tests

**1774 passed, 2 skipped.** `docker_image_inventory.py`, `docker.py`, `queue_uploader.py` and `log_uploader.py` are at **100% line and branch** coverage; the 2 remaining partials in `packages.py` are pre-existing, in the untouched host rpm/apk parsers.

Worth noting: the original "100% coverage" was **line** coverage -- the Makefile does not pass `--cov-branch`. Adding branch coverage surfaced 9 genuinely untested paths, including one where `_docker_socket_url` is assigned before `_probe_all()`: move that line and every agent dies at boot with `AttributeError`, while all five probe tests stay green (they patch `_can_access_docker` out).

The suite is green at **every one of the 7 commits** (verified by checking out each one), so `git bisect` works.

SDK-boundary mocking is deliberate for CI: the Linux jobs only build and lint, and the Windows job runs the full suite on a runner that cannot run Linux containers, so an automated E2E would be permanently skipped. The manual live run above covers the three beliefs the mock encodes (symlink entry + absolute `linkTarget`, 404 on a missing path, and the DB path constants) -- re-run it on any change to `_read_container_file` / `_get_archive` / the path constants.

## Follow-ups recorded in TODOS.md

- **P2** `/image_packages` merge-vs-replace semantics and the `ecosystem` gap in `packages_hash` -- both block the language-ecosystem phase.
- **P3** RPM extraction (phase 1.5); RPM images currently report `unsupported_distro` with a reason.
- **P4** a `package_cap` fixture scenario (declared in the enum, no scenario, since a 2001-package fixture would bloat the vendored file).

## Review decisions applied

The pre-landing review surfaced four judgment calls; all four were decided and implemented in this branch:

- **Rootless divergence** -- the XDG fallback was added to the *collector* (`docker.from_env()` has none), rather than removed from the probe. So the documented four-step order is now true of both, and rootless works without `DOCKER_HOST` when the agent runs as the rootless user.
- **Retry backoff** -- per-digest exponential backoff plus a give-up after `max_attempts`. Without it a failing digest is re-selected every tick and, because build runs before send, re-extracts in full each time. The give-up set is in-memory so a long outage cannot permanently blind an image; a queue-full shed releases the slot *without* counting an attempt, since backpressure is not a failed attempt.
- **Endpoint scheme validation** -- a non-unix endpoint is still assumed reachable (the Ceph light-probe precedent: a transient network failure must not disable a collector for a whole reprobe window), but the scheme is checked against what docker-py can actually dial, so a malformed value cannot report Docker as available on a host with no Docker.
- **Uploader duplication** -- extracted into a shared `QueueUploader` base rather than deferred; `LogUploader` now subclasses it with identical control flow.

## CI note

Neither workflow triggers on `pull_request` (they run on `main`, `release/**`, `feature/**`, `fix/**` and tags), so `issue-101` gets no automatic checks. Both were dispatched manually against this branch.

That mattered: the Windows job, which runs the **full** pytest suite, failed on `os.path.join(xdg, "docker.sock")` resolving through `ntpath` to `/run/user/1000\docker.sock` -- the same shape as the AF_UNIX break in #100. Fixed by building the path POSIX-style, which is correct on every platform since a Docker unix socket path has no Windows form. Worth considering adding `pull_request` to those triggers so this class of break is caught without a manual dispatch.


spuyet:
Remplacee par #111 : le renommage de la branche `issue-101` -> `feature/issue-101` (pour que la CI se declenche automatiquement, `feature/**` etant dans les triggers) a supprime la head ref, ce qui a ferme cette PR. Meme contenu, memes 8 commits, meme SHA de tete.

PR #117, opened 2026-07-31
perf(ci): cut Linux build time by baking the build env into the builder images

spuyet:
## What & why

CI Linux builds spent most of their time rebuilding the **same build environment from scratch on every job**. Measured from the CI logs (`build-synology-amd64` py2exe run, ~267s):

| phase | duration |
|---|---|
| `pip install poetry` (the tool) | ~120s |
| `poetry install` (deps) | ~122s |
| libpython3.10.so from source | ~50s |
| PyInstaller | ~16s |

~240s of that was reinstalling poetry + deps stone-cold every job (fresh venv on the mounted volume, and `poetry cache clear --all` wiped the cache first). This PR bakes the whole build environment into the builder image so each job only does the work that actually changes.

## Commits (each validated green on its own push)

1. **`perf(ci): pre-bake libpython3.10.so into the builder images`** — moves the CPython shared-lib compile from every py2exe run into the image (`Dockerfile`/`Dockerfile.arm`), matching the existing libtirpc/libvirt pattern. py2exe.sh's existing `[ ! -f ]` guard skips it when present.
2. **`fix(ci): pin builder image by commit SHA for Synology reuse`** — `build-synology-*` reused the builder image via the global `:latest` tag, which any concurrent run overwrites. It now pulls `:${{ github.sha }}` (build-amd64/arm64 push that tag alongside `:latest`), so Synology always gets this commit's image. This was a latent **correctness** bug: a concurrent build could hand Synology a foreign-commit toolchain.
3. **`perf(ci): bake the Python build env into the builder images`** — bakes `/opt/venv` (poetry + all locked deps, `--no-root`, libvirt-python compiled once) into the image, keyed on `pyproject.toml`+`poetry.lock`. py2exe.sh's fast path activates it and installs only the root project editable from the mounted source (`pip install -e . --no-deps`) — live code + correct `--copy-metadata` version, zero dep reinstall. The full from-scratch flow is preserved as a fallback for local/non-image builds.

## Measured results

- **`build-synology-amd64`: 387–432s → 66s** (py2exe body ~23s). Proven and stable across runs — Synology reuses the SHA-pinned **image**, so it's immune to buildcache state.
- **`build-synology-arm64`: 215–218s → 58–74s**.
- **`build-amd64` warm ≈ ~130s** (491s − the 358s of now-cacheable libpython+venv layers). The dep-bake layer is ~306s, built once and served from the registry buildcache thereafter.
- **Correctness validated**: the full distro matrix (centos7 / rocky 8/9/10 / ubuntu / debian / fedora / alpine, amd64+arm64) and the SELinux VM test **pass on the baked-venv binaries**; the `--version != 0.0.0+unknown` smoke check holds.

## Note on the buildcache

The shared `-builder-*:buildcache` ref is contended across branches (same family as the `:latest` race fixed in commit 2, but **benign** — a cache miss just rebuilds the layer). During this feature branch's life, main's back-to-back releases (v1.14.0→v1.14.4, same day) repeatedly overwrote the shared cache, so `build-amd64` rebuilt the new layers instead of cache-hitting them, which is why a clean warm measurement wasn't possible here. Post-merge, main owns its own buildcache and cache-hits the layers in the normal case. Not fixing it here on purpose: branch-scoping the cache would make every feature branch's first build cold.

## Risk / rollback

Build-pipeline only; no agent runtime change, no version bump. The `/opt/venv` fast path is guarded (`[ -d /opt/venv ]`) with the original build flow preserved as fallback. Revert is a clean 3-commit revert.


spuyet:
**Ship validation (merged state `fbec803`)** — branch brought current with `main` @ 1.14.4 (memcached/haproxy/rabbitmq collectors merged, no conflicts, `poetry.lock` unchanged). No version bump: CI-only change.

Both workflows green on the merged state:
- **Build and Release**: all jobs pass — 14-job distro matrix (centos7 -> rocky10, ubuntu/debian/fedora/alpine, amd64+arm64) + SELinux VM + builds. Fast path fires (`Using pre-baked build venv`), libpython skipped.
- **Windows Build**: full pytest suite green.

Measured: build-synology-amd64 66s (was 387-432s), build-synology-arm64 73s (was 215-218s). Merge-ready.

PR #120, opened 2026-08-06
v1.15.1 fix: honour the server's rescan_images directive so an api_error image is no longer unscannable forever (server #676)

spuyet:
## Summary

Agent half of **Five-Nines-io/fivenines_server#676**. The server half ([fivenines_server#714](https://github.com/Five-Nines-io/fivenines_server/pull/714), merged 2026-08-04) already emits a bounded, TOP-LEVEL `config["rescan_images"]` listing `api_error` digests the host runs. **Nothing read it**, so the directive was inert and the bug it exists to fix was still live in production.

**Why an image got stuck:** an `api_error` is reported INSIDE the `/image_packages` 200 -- the failure is in the payload, not the status code -- so `mark_done` recorded the digest as done. Digests are immutable, so nothing on the agent side could ever re-offer one. A busy daemon or a driver hiccup during extraction meant the image read "not scannable" **forever**.

## Changes

- **`ImageInventoryCoordinator.reset(image_ids)`** -- drops digests from the done set and from `gave_up`, persists the new done set (a restart must not resurrect it), skips in-flight digests (`select_jobs` would skip them anyway, and clearing the slot would let one extraction be enqueued twice), and re-writes state only when something was actually removed.
- **`apply_rescan_requests(coordinator, config)`** -- the untrusted-input glue, mirroring `log_capture.evaluate_and_enqueue`. Anything that is not a list of non-empty strings is ignored rather than raised on; the list is capped at `MAX_RESCAN_IMAGES` (10, vs the server's own 3/tick) so a malformed config cannot re-open the whole done set.
- Wired into `Agent._handle_image_inventory` **before** selection, so a re-opened digest is offered on the same tick, and only while `image_inventory` is enabled -- the server must not mutate agent state the operator turned off.

## The anti-treadmill guard (found by mutation-testing my own first draft)

My first version left the backoff alone and I wrote a test claiming that was the guard. **The mutation passed** -- the test was vacuous, and digging into why exposed a real design hole rather than just a bad assertion:

`mark_done` clears the failure history and `mark_failed` pops `_next_retry` when it gives up, so a done-or-given-up digest never carries a live backoff. That meant a given-up digest, once re-opened, was **immediately** eligible. And when the POST itself is what fails, the server never receives a result, never restamps its 6h backoff, and re-requests the digest **every tick** -- so the loop was: re-open -> full archive fetch + tar parse -> fail -> give up -> re-open, once per tick, forever.

`reset()` now re-arms `_next_retry` through a shared `_retry_delay()` (extracted from `mark_failed` so the two cadences cannot drift):

- **no failure history** (the main #676 case -- `mark_done` cleared it) -> re-offered on the same tick, no delay
- **previously given up** -> re-opened but paced by the same exponential ladder, climbing to the 3600s cap

## Tests

**2033 passed, 2 skipped**; `docker_image_inventory.py` at **100%** coverage (`agent.py` unchanged at its pre-existing 94%). +21 examples covering: re-open of a done digest, persistence across a restart, `gave_up` clearing, the ladder on both paths, in-flight skip, unknown-digest no-op, duplicate/empty input, no disk write when nothing was removed, the malformed-config matrix, the cap, and the end-to-end Agent hook.

**Mutation-verified:**
- removing the `apply_rescan_requests` call from `agent.py` -> fails `test_handle_inventory_honours_rescan_images` by name
- making `reset()` clear the backoff -> fails `test_reset_of_a_given_up_digest_stays_on_the_retry_ladder` **and** `test_repeated_reset_of_a_failing_digest_cannot_extract_every_tick`

A mirror test (`test_handle_inventory_without_rescan_leaves_a_done_digest_alone`) pins that the directive, not the ordinary selection path, is what re-opens the digest.

## Contract fixture

`tests/fixtures/docker_image_inventory_contract_payload.json` reconciled **byte-for-byte** with the server copy (verified with `diff`): adds `config.rescan_images` + its note, and rewrites `errors_contract.type.api_error` to describe the now-closed loop. `agent_min_version` stays **1.14.0** -- that gates the image-inventory feature itself, and the rescan key is safe to send to any agent (an unknown top-level key is ignored), so it needs no new version floor.

The identical fixture edit ships in a companion server PR; the two must land together to keep the contract byte-identical.

## Deploy

Server-first is already done. Once 1.15.1 rolls out, images stuck in `api_error` self-heal on the next tick that carries the directive. No config change, no new permission -- it reuses the socket the Docker collector already has.


spuyet:
Superseded by #121, which ships instead.

#121 is an independent hardened rewrite (not stacked on this branch — `git merge-base --is-ancestor` confirms no shared history). I audited it and it is strictly better than this PR:

- **It found the assumption this PR got wrong.** I reasoned that the server paces re-extraction via its 6h backoff. But the `Synchronizer` replays the last `/collect` config during an outage, so the directive is level-triggered from a *cached* config with the server not involved — and if the POST itself fails the server never restamps `inventory_received_at`, so it re-requests every tick anyway. This PR only floored the `gave_up` path; #121 floors **every** re-open per digest (`rescan_min_interval`, 1h).
- **The input cap here bounded what was *honoured*, not what was *scanned*** — a huge array was still fully iterated on the watchdog-bounded collection loop. #121 adds `MAX_RESCAN_SCANNED`.
- **The persist race I flagged is fixed, and better than I planned**: #121 re-reads `_done` under `_lock` *and* writes via temp-file + `os.replace` under a dedicated lock, closing the torn-file hazard I had not considered.

Independently verified on #121: 2040 passed, 100% line **and** branch coverage, and I mutation-tested its two headline guards myself (removing the treadmill floor fails 2 tests by name; removing `os.replace` fails 1).

Server-side fixture reconcile Five-Nines-io/fivenines_server#720 is synced byte-for-byte to **#121**, not to this branch.

PR #128, opened 2026-08-14
v1.16.0 feat: WireGuard peer health + Tailscale key expiry collectors (server #508)

### LINKED ISSUE #127: WireGuard peer health + Tailscale key expiry collectors (server #508 companion)
Companion to **server [#508](https://github.com/Five-Nines-io/fivenines_server/issues/508)** (server half merged in [fivenines_server#756](https://github.com/Five-Nines-io/fivenines_server/pull/756)).

The server side is **shipped and inert**: `wireguard_enabled` / `tailscale_enabled` default to false, there is **no `FEATURES_SUPPORTED_VERSIONS` gate**, and nothing happens until the agent starts sending `data["wireguard"]` / `data["tailscale"]`. Deploy order is server-first, which is already done, so this can ship whenever.

MSPs run WireGuard to every customer site, so a dead tunnel means blind monitoring of *everything behind it*. That is the headline: `wireguard_peer_stale` fires per peer.

## The contract is already frozen

The server ships `spec/fixtures/vpn_contract_payload.json` with 6 scenarios (`healthy_and_stale`, `key_expiring`, `needs_login`, `no_expiry`, `zero_peers`, `collection_failure`). **Vendor it byte-for-byte** into `tests/fixtures/vpn_contract_payload.json` and assert it end-to-end here (mock only the `wg show all dump` / `tailscale status --json` subprocess, feed each scenario's `raw` back in, and require the collector to produce `scenario["payload"]`). Once vendored, the **agent repo is the source of truth** and the two copies change only in lockstep - same discipline as the ceph / systemd / haproxy fixtures.

The server's `spec/requests/api_collect_vpn_spec.rb` posts each scenario through `/collect`, so a shape mismatch is caught on both sides.

## Config the server already emits

Both keys are **TOP-LEVEL and plain booleans** (no splat hazard - an agent without the collectors just ignores them):

```json
{ "wireguard": true, "tailscale": true }
```

- `config["wireguard"]` is **Linux-only** - it is in `Host::WINDOWS_OMIT_CONFIG_KEYS`, so it is stripped for Windows agents (`wg` and its dump format are the Linux/BSD userspace tools).
- `config["tailscale"]` is **cross-OS** and never stripped (tailscaled and `tailscale status --json` are identical on Linux/Windows/macOS).

Neither collector takes parameters: the commands are fixed, and peer aliases come from the local wg-quick configs, not from server state.

## WireGuard payload

```json
{
  "interfaces": [
    { "name": "wg0", "listen_port": 51820, "peer_count": 4 }
  ],
  "peers": [
    {
      "interface": "wg0",
      "public_key": "xTIBA5rboUvnH4htodjb6e697QjLERt1NAB4mZqp8Dg=",
      "name": "office-router",
      "endpoint": "203.0.113.5:51820",
      "allowed_ips": "10.100.0.2/32",
      "last_handshake_age_seconds": 42,
      "rx_bytes": 918273645,
      "tx_bytes": 546372819,
      "persistent_keepalive": 25
    }
  ]
}
```

### The five things that do not get guessed right

**1. Strip the secrets.** `wg show all dump` prints the interface **PRIVATE KEY** as the first field of each interface line, and each peer's **PRESHARED KEY**. Both must be stripped and never transmitted. The server has no column for either and would drop them. Only the **public** key travels - it is public by construction and is the peer's durable identity.

**2. `null` and `[]` mean opposite things.**
- `data["wireguard"] = null` = **collection failure** (`wg` missing, no `CAP_NET_ADMIN`, timeout). The server's dispatch gate is `is_a?(Hash)`, so it skips ingestion entirely: no prune, no auto-resolve.
- `{"peers": []}` = `wg` succeeded and there are genuinely **zero** peers: prune-all.
- A non-empty `peers` array is the **FULL current set**, so a peer absent from it is genuinely gone and does vanish-prune.

Reporting `[]` on a privilege failure would mass-false-resolve every open `wireguard_peer_stale` incident. That is the failure mode this split exists to prevent.

**3. On a zero-peer tick, `interfaces` must still list every configured interface.** It is the only source of names for the per-interface rollup gauges when there are no peer rows to derive them from. Sending `interfaces: []` alongside `peers: []` makes `wireguard_peers_total` / `_active` **vanish from the chart** instead of reading an honest 0. See the `zero_peers` scenario.

**4. Send an AGE, not a timestamp.** `last_handshake_age_seconds` is computed agent-side from the agent's own clock; the server anchors it to its own `received_at` (the mqtt/qemu/systemd clock-discipline rule). `null` when `wg` reports `latest-handshake=0`, i.e. the peer has never completed a handshake since the interface came up - **not** 0.

**5. `persistent_keepalive` is load-bearing.** WireGuard is silent by design, so a peer *without* keepalive can be idle-but-healthy with an arbitrarily old handshake. The `wireguard_peer_stale` trigger watches only keepalive-configured peers by default for exactly that reason. Send the interval in seconds, or `null`/`0` when off.

### Field notes

- `rx_bytes` / `tx_bytes` are **cumulative**; they reset when the interface is recreated (`rate()` handles that natively).
- `name` is the operator alias parsed from the wg-quick config - the `# Name = ...` comment immediately preceding the `[Peer]` block, the de-facto convention. `null` when absent; the server then renders a short key fingerprint.
- `endpoint` is `null` for a roaming client that has never dialled in.
- `interfaces[].listen_port` and `interfaces[].peer_count` are **RESERVED**: transmitted but not read by the server today. In particular there is deliberately **no** cross-check of `peer_count` against the peers array - do not assume payload drift is guarded by one.

### Privilege

`wg show all dump` needs root or `CAP_NET_ADMIN`. Grant **`AmbientCapabilities=CAP_NET_ADMIN`** in the packaged systemd unit rather than running the agent as root. Document it, and make sure the failure path reports `null` (point 2).

## Tailscale payload

```json
{
  "backend_state": "Running",
  "self": {
    "hostname": "edge-01",
    "key_expiry": "2026-09-12T08:00:00Z",
    "online": true
  },
  "peers_total": 12,
  "peers_online": 11
}
```

This half exists for **one failure mode**: an expired node key silently drops the machine off the tailnet. tailscaled stays alive, flips to `NeedsLogin`, and nothing on the box logs an error while every tailnet-only service stops answering.

- `backend_state` is tailscale's own `BackendState` **verbatim as a string** (`Running` / `Stopped` / `NeedsLogin` / `NeedsMachineAuth` / `Starting` / `NoState`). Never an enum - the vocabulary is tailscale's and they extend it.
- `self.key_expiry` is an ISO-8601 UTC instant, or **`null` when key expiry is DISABLED** for the node (a legitimate admin-console setting). `null` means "never expires", not "unknown" - the server emits no sample at all in that case rather than charting 0 days left.
- `peers_total` / `peers_online` are **rollups only**. Do not send per-peer rows: every node sees every tailnet peer, so per-peer series would cost `hosts x peers` cardinality across a fleet (the systemd #562 precedent).
- `self.online` is **RESERVED** - transmitted but not read today. It is the coordination server's view, deliberately *not* the same signal as the server's `tailscale_up` (derived from `backend_state`, the local daemon's view), so shipping it must not be read as coverage of that.
- Same `null` rule as WireGuard: `data["tailscale"] = null` means the CLI/daemon could not be reached at all, and the server preserves the last-known denormalized block rather than blanking it. "Off the tailnet" is a **successful** read carrying `Stopped` / `NeedsLogin`, never `null`.

## Acceptance

- [ ] `wireguard` collector, Linux-only, gated on `config["wireguard"]`
- [ ] Secrets (interface private key, peer preshared key) stripped and never transmitted
- [ ] Privilege failure / missing `wg` / timeout reports `null`, never `[]`
- [ ] `interfaces` populated even on a zero-peer tick
- [ ] `tailscale` collector, cross-OS, gated on `config["tailscale"]`, rollups only
- [ ] `AmbientCapabilities=CAP_NET_ADMIN` in the packaged unit + docs
- [ ] `tests/fixtures/vpn_contract_payload.json` vendored byte-identical from the server and asserted end-to-end
- [ ] Version bumped to **1.16.0** (the number the server's contract fixture documents for the lockstep; nothing on the server gates on it)

spuyet:
## Summary

Agent half of **server #508** (server merged in Five-Nines-io/fivenines_server#756). Deploy order is server-first and the server is already shipped and inert (`wireguard_enabled` / `tailscale_enabled` default false, no `FEATURES_SUPPORTED_VERSIONS` gate), so this can land whenever.

MSPs run WireGuard to every customer site, so a dead tunnel means blind monitoring of *everything behind it*. That is the headline: `wireguard_peer_stale` fires per peer.

**Collectors**
- `fivenines_agent/wireguard.py` (8f6eaec) — one `wg show all dump` per tick becomes the full current set of interfaces and peers.
- `fivenines_agent/tailscale.py` (b3f87cd) — one `tailscale status --json` per tick becomes `backend_state`, node key expiry, and tailnet rollups.

**Wiring** (d0d5f3a) — both registered in the `COLLECTORS` registry as TOP-LEVEL plain booleans with `pass_kwargs=False`, so a future dict-valued config key is ignored rather than splatted into a `TypeError` (the `image_inventory` hazard). Neither is capability-gated: a privilege failure must surface as `data["wireguard"] = null`, not a missing key. Both added to `_DRY_RUN_CONFIG` so `--dry-run` actually shows the payload.

**Contract** (ddccba7) — `tests/fixtures/vpn_contract_payload.json` vendored from the server, all 6 scenarios round-tripped end-to-end.

**Packaging** (1757084) — `AmbientCapabilities=CAP_NET_ADMIN` in the systemd unit.

### The decisions that are not obvious

**1. ANY stderr from `wg show all dump` means `null`, not just a non-zero exit.**
Without `CAP_NET_ADMIN`, `wg` can still enumerate interface *names* over rtnetlink but cannot read them, so it prints a per-interface EPERM line and **can still exit 0 having emitted a partial dump**. The server reads a non-empty `peers` array as the FULL current set, so shipping that partial view would vanish-prune every peer we could not read and mass-false-resolve live `wireguard_peer_stale` incidents. A data gap is recoverable; a false prune is not. The safety valve: that path logs at `error` level, and `debug.py` buffers error logs into backend telemetry, so a permanently-null collector reports itself every tick instead of going silently blind.

**2. A negative handshake age reports `null`, never `0`.** Found by the adversarial pass, and it is the same failure class the feature exists to prevent. If the host clock steps backward (NTP correcting a skewed host, a VM restored from a snapshot), `wg`'s `latest-handshake` lands in the future. The server computes `last_handshake_at = received_at - age` and keeps `[candidate, prior_at].max`, so an age of `0` **wins against the real, older anchor** and silently resolves the incident on a tunnel that is actually dead. `null` means "no reading this tick", the true anchor survives, and the age keeps growing. Nonsense data must never outrank real data.

**3. Peer aliases are tie-broken on the block's own `PublicKey`.** The `# Name = ...` comment convention puts the alias immediately *before* `[Peer]`, which is textually identical to "trailing the previous block" once blank lines are dropped. A comment seen before this block's `PublicKey` names this peer; one seen after it heads the NEXT block. Getting this backwards mislabels every peer in the file by one.

**4. No agent-side cap on `peers`, deliberately.** A truncated array is indistinguishable from a shrunken one, so an agent-side cap would vanish-prune everything past the cut. The bound belongs server-side, where `Ingesters::Wireguard` already caps, logs which peers went unmonitored, and sets `safe_to_prune? == false` for that tick. Recorded as a DO-NOT in the collector docstring.

**5. Secrets never travel.** The dump prints the interface private key and each peer's preshared key; both columns are indexed past and never read. Only the peer public key ships. The fixture deliberately carries real-shaped fake secrets so the round-trip *proves* stripping rather than assuming it.

**6. Tailscale branches on the document, not the exit code.** `tailscale status` exits 1 when the node is not up, but its `--json` branch returns before that check, so a logged-out daemon prints a full document and exits 0. A parseable document carrying `NeedsLogin` is a SUCCESSFUL read — that is the signal, not a failure.

### Cross-checked against the merged server

Every field `Ingesters::Wireguard` and `Ingesters::Tailscale` read matches what the collectors emit. `interfaces[].listen_port`, `interfaces[].peer_count`, and `self.online` confirmed to have no reader anywhere in the server — genuinely RESERVED, as the contract documents.

## Test Coverage

```
COVERAGE: 100%  |  wireguard.py 146/146 stmts, 54/54 branches
                |  tailscale.py  89/89 stmts,  34/34 branches
QUALITY: ***:14  |  GAPS: 0
```

Ran with `--cov-branch`, not just line coverage: a prior learning on this repo records that `make test` gates line coverage only, and branch gaps hide behind it.

Covered: every `null` path (binary absent, spawn error, timeout, non-zero exit with and without stderr, zero-exit-with-stderr, unparseable output), honest-empty vs collection-failure, zero-peer interfaces, never-handshaked and future-dated handshakes, all six `persistent_keepalive` spellings, roaming peers with no endpoint, malformed dump lines, peers on undeclared interfaces, all three alias-comment placements, unreadable configs, path-traversal refusal, RFC3339 with `Z`/`z`/offset/naive/9-digit nanos/zero-time/unparseable, absent `Self`, null and non-dict `Peer` maps, and a secret sweep across all 6 contract scenarios.

Tests: 62 -> 65 files (+3), 2132 passing.

## Pre-Landing Review

5 findings, 0 critical. 4 auto-fixed:

- `agent.py` — enum completeness: new config keys missing from `_DRY_RUN_CONFIG`, so `--dry-run` would never exercise the collectors. Added both plus a regression test, matching the docker/zfs/systemd precedent already in `test_agent_dry_run.py`.
- `wireguard.py` — `except OSError` missed `open()`'s `ValueError` on an embedded null byte. Widened; aliases are best-effort by contract, so a read failure must degrade to "no alias" rather than raise the whole tick to null.
- `wireguard.py` — documented that the absent peers cap is deliberate, so a future maintainer does not "fix" it into a vanish-prune.
- `test_vpn_contract.py` — `wireguard.subprocess` and `tailscale.subprocess` are the same module object, so correctness rested on implicit call ordering. The fake now asserts `argv[0]`.

1 escalated (not fixable in this repo) — see Follow-up.

## Adversarial Review

**Reduced coverage, stated plainly: no independent model reviewed this diff.** Codex returned 401 (`refresh_token_reused` — needs `codex login`). The adversarial pass ran inline instead, which means the author reviewed their own code. It did find one real bug (the negative-age clamp above), but self-review is not independence. Worth a second pair of eyes on the `null`-vs-`0`-vs-`[]` decisions specifically.

## Scope Drift

`Scope Check: CLEAN` — 2 collectors, registry wiring, vendored fixture, 3 test files, CAP_NET_ADMIN, docs, version. No unrelated files touched.

## Plan Completion

Audited against the issue #127 acceptance checklist (no gstack plan file). **8/8 DONE**, 0 partial, 0 deferred: Linux-only config-gated wireguard collector; secrets stripped; privilege failure reports null never `[]`; interfaces populated on a zero-peer tick; cross-OS rollups-only tailscale collector; CAP_NET_ADMIN in the unit plus docs; fixture vendored and asserted end-to-end; version 1.16.0.

## Verification Results

Skipped — this is a headless agent with no web surface, so there are no browser-verifiable steps.

## TODOS

No existing items completed. One added:
- **P1: Reconcile the server's copy of `vpn_contract_payload.json`** (see Follow-up).

## Documentation

- `README.md` — new **VPN Monitoring** section: both collectors, the keepalive-vs-idle-peer rule that decides which peers are watched, secret handling, and honestly that OpenRC/Alpine and user-level installs cannot grant `CAP_NET_ADMIN`, so WireGuard reports null there rather than pretending to work. Permissions table gains a row per collector.
- `CLAUDE.md` — new **VPN** collector entry and the top-level config-key contract.

## Follow-up required (cross-repo lockstep)

The server's `spec/fixtures/vpn_contract_payload.json` still carries the placeholder `raw` strings it was authored with. **Nothing is broken today** — every key the server actually reads (`agent_min_version`, the five contract docs, and every scenario's `description` / `config` / `payload`) is byte-identical, verified programmatically. But the agent copy is now the source of truth and the two have drifted in the agent-side inputs.

Fix is a one-file copy; the server's `api_collect_vpn_spec.rb` only reads `scenarios.*.payload`, so it cannot break that suite.

## Test plan

- [x] Full suite: 2132 passed, 5 skipped
- [x] Both new modules 100% line AND branch coverage
- [x] All 6 contract scenarios round-trip through the real collectors
- [x] No interface private key or peer preshared key reaches any payload (asserted across all scenarios)
- [x] ASCII gate clean, flake8 clean
- [x] Every scenario payload serializes with `allow_nan=False`
- [ ] CI on this branch (why it was renamed off `issue-127`: neither workflow triggers on `pull_request` or `issue-*`, and `windows.yml` is the only CI that runs pytest)

Closes #127


spuyet:
## Independent adversarial review (Codex) — 2 rounds

The PR body's caveat about "no independent model reviewed this diff" is now resolved. Codex was re-authenticated and ran twice. It initially returned **reject**, and it was right: the null-vs-empty discipline was enforced at the *subprocess* boundary (exit code, stderr) but not at the *parse* boundary, leaving three ways to report a failure as healthy data.

### Round 1 — 4 findings, verdict `reject`

**1. Malformed dump rows were skipped, not fatal.** `_parse_dump` quietly `continue`d past bad field counts, blank names, and peers whose interface row never appeared. The server reads a non-empty `peers` array as the FULL current set, so a short list vanish-prunes every dropped peer and resolves its open incident.

The realistic trigger is not a corrupt byte — it's a **format change**. If wireguard-tools ever appends a column, every peer line stops matching at once, silently turning every WireGuard host in the fleet into "zero peers, prune everything" on the same day. Now raises `_MalformedDump` and the tick reports `null`.

**2. Lenient numeric defaults fabricated healthy values.** The dangerous one is keepalive: an unparseable value became `None`, which means "the operator turned keepalive off" and drops the peer out of the server's default stale-watch set — silently unwatching a monitored tunnel while still reporting a successful full-set read. Transfer counters coerced to `0` read as a counter reset. `_strict_int` now raises; only wg's real off-sentinels map to `None`.

**3. Ambient `CAP_NET_ADMIN` on the whole agent, every install.** Left as-is — a decided trade-off from #127, documented in the unit with a "delete this line if you don't monitor WireGuard" escape hatch. Isolating the capability to just the `wg` invocation isn't available without a setcap'd wrapper or a privileged helper, neither of which the agent should install.

**4. Tailscale accepted a document with no `Self`** and published `key_expiry: null`, which contractually means "expiry is DISABLED", not "unknown". A shape change would switch off expiry monitoring exactly on nodes whose keys are expiring. `Self` is now required.

### Round 2 — verification, verdict `accept`

Codex re-reviewed `c13cc2a` adversarially and confirmed **1, 2, 4 CLOSED**, citing the specific lines, and explicitly checked for over-correction: *"I do not see a convincing new WireGuard regression from the stricter parser... that may null a future changed-but-still-readable wg dump, but that is the safer side of the contract because the alternative is authoritative partial-set pruning."*

It found two narrower residuals of the same class, closed in `01539d6` rather than queued:

- `Self.KeyExpiry` **present but not a string** was collapsed to `None` (= "expiry disabled"). Absent/null still means disabled (the legitimate Go `omitempty` reading); present-but-wrong-type now fails the tick.
- `Peer` **present but not an object** was downgraded to `{}`, reporting an empty tailnet the node cannot see. Absent/null still means 0/0 (the logged-out reading, per the `needs_login` scenario); present-but-wrong-type now fails the tick.

### State after the fixes

- **2144 passed**, 5 skipped (was 2132 — 12 new tests for the strict paths)
- Both modules **100% line and branch** coverage
- Removed one now-dead guard (`if interface in by_name`) that the new raise makes unreachable
- Prior CI run was green on both workflows before these commits; re-running now

PR #141, opened 2026-09-07
v1.17.4 sec: harden against hostile config, local attackers, and DoS

spuyet:
## Summary

> **Stacked PR** — base is `perf/agent-tick-overhead` (#140). Merge #140 first; this PR carries only the security work.

Deep security-hardening pass from a full review of the agent under this threat model, in priority order: a compromised/malicious fivenines API pushing hostile config fleet-wide; a local unprivileged attacker on the monitored host; network attackers; resource-exhaustion DoS. Every Python fix ships with a regression test.

**Hostile-config RCE pivot closed (redis)**
- The Redis collector spoke the inline protocol and joined commands with CRLF, so a config-supplied password of `x\r\nCONFIG SET dir ...` executed arbitrary commands on the local Redis — the classic `CONFIG SET` → RCE pivot, reachable by whoever controls `/collect` responses. AUTH/INFO/QUIT now use RESP array framing: the password travels as one length-prefixed bulk string (byte length, pinned by a multibyte test), so embedded CRLF is pure data. Bonus: passwords containing spaces now work (inline framing silently broke them).
- The INFO reply read also gains a byte cap + wall-clock deadline (a trickling local listener on the config-supplied port could stall the watchdog-bounded loop; the 5s socket timeout is per-recv inactivity).

**SSRF / unbounded-read hardening (server-pushed URLs)**
- New shared `http_body.read_capped_body()`: streamed reads under a byte cap AND a wall-clock deadline (requests' `timeout` is per-socket-op — a 1-byte-per-4s trickle never trips it), redirects refused. Wired into `tsdb` (8MB/15s), `php_fpm` (1MB/5s), `rabbitmq` (16MB/20s), matching the `inference_metrics`/`haproxy` posture the repo already had. Streamed responses are closed on every non-parse path (an unread body keeps the connection checked out of the keep-alive pool).
- Ping: a hostile/fat-fingered `ping` map could stretch a tick past WatchdogSec=90 (sequential 5s blocks) and restart-loop the fleet — now capped at 10 targets/tick with per-tick rotation (no permanently starved tail), 200-char field clamps, a 30s loop deadline, and a non-dict `ping` value (e.g. `ping: true`) yields zero targets instead of an AttributeError that exited the agent into a `Restart=always` crash loop.

**Secrets / redaction**
- `redact()` now catches cephx key material: a cephx secret is 38 base64 chars + `==`, which slipped under the old 40-char blob threshold in every shape ceph actually emits (`key = AQ...` keyring lines, decode errors). Threshold lowered to 38, a key-labeled rule added, and the blob rule moved to position 2 — which also kills quadratic regex backtracking in the email and conn-string rules (measured: seconds of CPU per 100KB token, on the watchdog-bounded loop; the conn-string rule additionally gains its load-bearing `\b` anchor).
- ceph CLI stderr is redacted before it rides the payload (config-supplied `conf`/`keyring` paths mean ceph may quote lines of whatever file it was pointed at into its parse errors); classification still runs on the raw stderr, pinned by a mutation-killing test, and the redact input is prefix-bounded.

**Local attacker / file permissions**
- TOKEN swap, `last_capture_id`, and `image_inventory_done` are created 0600 umask-independently, with `fchmod` healing pre-existing group/world-readable files. A readable TOKEN lets a local user POST `get_config` and receive every service credential the config carries.

**systemd unit sandbox (the careful subset)**
- `UMask=0077`, `PrivateTmp=true`, `ProtectHome=read-only` — mount-namespace/attribute directives only. The unit now documents the trap that almost shipped here: on a non-root unit, EVERY seccomp-backed directive (`ProtectKernelTunables`, `RestrictRealtime`, `LockPersonality`, ...) makes systemd implicitly enable `NoNewPrivileges`, which breaks every `sudo -n smartctl/mdadm/fail2ban-client` spawn — SMART/RAID/fail2ban monitoring would have died silently fleet-wide on systemd >= 228 (caught by the review, verified against systemd's `context_has_no_new_privileges()`). Caveats documented in the unit: `/tmp` DB sockets under PrivateTmp, and the 226/NAMESPACE failure on namespace-less kernels, both with the drop-in escape hatch.

**Supply chain**
- `fivenines_setup.sh` / `fivenines_update.sh`: the GitHub fallback URL pointed at a nonexistent repo name (`five_nines_agent`) — every fallback download 404ed. Fixed to `fivenines_agent`.

**Reviewed and deliberately deferred (documented)**
- `apache.py` / `nginx.py` / `caddy.py` still use plain `requests.get` (same server-pushed-URL exposure class) — follow-up to bring them onto `http_body`.
- Release-artifact checksum publication + installer verification needs release-pipeline changes; out of scope for an agent-repo-only PR.
- `verify_ssl: false` config paths kept (explicit operator opt-in, matching the existing contract).

## Test Coverage

Tests: 2483 -> 2517 (+34 on this branch). Changed-path coverage: 24/25 (96%) per the audit, then the last gap (tsdb empty-chunk skip) closed in the review round; changed modules at 100% except pre-existing base misses. All 0600 tests pin the umask; the RESP framing test pins byte-length with a multibyte password; mutation-killing tests for ceph classify-on-raw and the redis length header.

<details>
<summary>Coverage audit diagram (Step 7)</summary>

```
ATTACK SCENARIO / CHANGED PATH                          SCORE
1. Hostile Redis password (CRLF injection)              *** x4 (frame bytes, order, injection, no-AUTH)
   + bounded INFO reply read (cap + deadline)           *** (review round)
2. Oversized/hostile ping map (watchdog DoS)            *** cap, rotation, warn-once, non-dict guard,
                                                            field clamps, loop deadline, wiring test
3. Redirecting/oversized/trickling endpoint (SSRF+DoS)
   tsdb / php_fpm / rabbitmq                            *** over-cap, redirects-refused, mid-stream reset,
                                                            close-on-every-path, empty-chunk skip
   shared http_body reader                              *** cap / deadline / close / propagation tests
4. Secrets in ceph stderr                               *** redacted + classify-on-RAW (mutation-killed)
                                                            + bounded redact input (timed test)
   cephx redaction (38-char blob, key= rule, ReDoS)     *** labeled + quoted shapes + timing pin
5. Permissive umask (local attacker)                    *** 3 sites, umask-pinned, fchmod healing,
                                                            os.open error paths re-pinned
NOT PYTEST-TESTABLE: systemd unit directives, setup-script URL fix (shell)
```

</details>

## Pre-Landing Review

5 specialists + red-team-style adversarial pass on the hardening diff itself: 18 findings (2 CRITICAL), all fixed or explicitly deferred. The two CRITICALs were in this branch's own first commit — both would have shipped broken hardening:

- **[CRITICAL, security]** the seccomp-class systemd directives implicitly enable `NoNewPrivileges` on a non-root unit → sudo-based collectors dead fleet-wide. Removed; the unit documents the trap.
- **[CRITICAL, security+performance, multi-confirmed]** `redact()` missed cephx key shapes entirely AND backtracked quadratically on long tokens (~350KB of stderr exceeds WatchdogSec in one call). Both fixed with timed + shape tests.
- Adversarial round 2: the `ping: true` crash-loop guard, the redis unbounded read, body-deadline decoupling, ping rotation. Codex passes could not run in this environment (bwrap needs unprivileged user namespaces) — missing outside-model coverage, stated rather than papered over.

## Design Review

No frontend files changed -- design review skipped.

## Eval Results

No prompt-related files changed -- evals skipped.

## Plan Completion

No plan file detected.

## TODOS

No TODO items completed in this PR.

## Documentation

Docs synced to the v1.17.4 hardening pass (diffed against `perf/agent-tick-overhead`), commit `d8a8ea6`.

- **README.md** (6 edits): systemd unit hardening note under Standard Installation (`UMask=0077`, `PrivateTmp=true`, `ProtectHome=read-only`; why `NoNewPrivileges`/seccomp/`ProtectSystem` are deliberately absent; the `/tmp`-socket and `226/NAMESPACE` caveats with the `systemctl edit` drop-in override). Redaction section: blob threshold corrected 40 -> 38 chars (the exact cephx key body length), Ceph keyring rule (`key = AQ...`) added to the mask list, and Ceph error-report stderr now covered. Redirect-refusal plus size-cap/wall-clock-bounded body reads documented for PHP-FPM (HTTP mode), RabbitMQ, and Prometheus/VictoriaMetrics (point configs at the final URL). Redis section: password now sent with length-prefixed protocol framing, so spaces/special characters work verbatim.
- **CLAUDE.md** (7 edits): new core component "Bounded HTTP Body Reads (`http_body.py`)" documenting the shared streamed/redirects-refused/byte-cap+deadline posture, per-caller budgets (tsdb 8 MB/15s, php_fpm 1 MB/5s, rabbitmq 16 MB/20s, redis raw-socket 1 MB/5s) and the must-use rule for new collectors; agent.py bullet for the ping bounds (MAX_PING_TARGETS=10/tick with rotation, 200-char field cap, 30s loop deadline, DNS-not-covered rationale); Synchronizer TOKEN-swap line plus a new "Config-Dir State Files" pattern (os.open 0600 + fchmod healing; used by TOKEN, last_capture_id, image_inventory_done); redis.py entry (RESP array framing kills the CRLF-injection/space-breaks-AUTH class) and rabbitmq.py entry (streamed+capped reads, error-path response closing); logs.py entry (order-sensitive `_REDACTIONS`, 38-char blob rule, cephx rule, ReDoS rationale); Installation Types guardrail: never add seccomp-backed directives or NoNewPrivileges to the unit (breaks every `sudo -n` collector fleet-wide).

### Documentation Debt

- `ping` targets: server-pushed config with no README section (pre-existing gap); the new caps are documented in CLAUDE.md only.
- `php_fpm.py`, `ceph.py`, `tsdb.py`, `caddy.py`, `nginx.py` have no entries in CLAUDE.md's collector list (pre-existing); only their new hardening posture is captured via the http_body component entry.
- TODOS.md and synology/README.md are not linked from README.md or CLAUDE.md (discoverability).

## Test plan
- [x] Full pytest suite passes (2517 passed, 5 skipped)
- [x] flake8 clean on all touched files; ASCII-only CI check clean


spuyet:
Superseded by #148.

This PR merged into `perf/agent-tick-overhead` 14 seconds after that branch itself landed on `main` as #140, so its merge commit is not an ancestor of `main` and none of this work shipped. #148 replays the same five commits on top of current `main` (resulting tree byte-identical, same tree hash `2be22f7`) and keeps version 1.17.4.

PR #158, opened 2026-09-21
v1.19.0 feat(proxmox): backups block -- per-guest last-backup age, vzdump tasks, jobs, not-backed-up (#156)

### LINKED ISSUE #156: Proxmox: backups block - per-guest last-backup age, vzdump task outcomes, backup jobs, not-backed-up guests (server companion of fivenines_server#1159, phase 2)
## Why

Server issue Five-Nines-io/fivenines_server#1159 is phase 1 (a heartbeat that can report failure). This is the agent half of **phase 2**: read, from the PVE API the agent already talks to, everything the server needs to answer *"has every VM and container had a successful backup in the last N hours, whatever the target (PBS, NFS, local)?"*, plus *"which guests are not in any backup job at all?"*.

Design context on the server side: `docs/designs/proxmox-backup-monitoring.md` (branch `proxmox-backup-history`). Customer context: an MSP evaluating us across several Proxmox/PBS environments, and other Proxmox users who raised backup monitoring before him.

Why PVE and not PBS: vzdump runs on the PVE node whatever the destination, so PVE sees 100% of backups, and the storage content listing goes through PVE even for a PBS datastore. PBS-native monitoring (verify/sync/prune/GC jobs) is a separate, later integration.

**No new credentials.** All four reads are covered by the role our setup guide already provisions (`pveum acl modify / --roles PVEAuditor`, server docs/proxmox.md:334):

| Read | Endpoint | Declared permission (pve-docs api-viewer) |
|---|---|---|
| Backup volumes | `GET /nodes/{node}/storage/{storage}/content?content=backup` | `Datastore.Audit` or `Datastore.AllocateSpace` on `/storage/{storage}` |
| vzdump task outcomes | `GET /nodes/{node}/tasks?typefilter=vzdump` | own tasks, or all with `Sys.Audit` on `/nodes/{node}` |
| Backup job definitions | `GET /cluster/backup` | `Sys.Audit` on `/` |
| Guests in no job | `GET /cluster/backup-info/not-backed-up` | `Sys.Audit` on `/` |

`PVEAuditor` is built in `pve-access-control` `AccessControl.pm` `create_roles` as the union of every category's `audit` privilege group, which contains `Sys.Audit` and `Datastore.Audit`. Documented evidence, not runtime proof: the first acceptance criterion below is one curl with the guide's token on a real PVE.

### 1. Per-guest latest backup, per storage (the core signal)

`GET /nodes/{node}/storage/{storage}/content?content=backup`, inside the existing `nodes -> storages` loop of `_collect_storage`. Return fields that matter: `volid`, `vmid`, `ctime` (epoch), `size` (fall back to `approximate-size`), `format`, `protected`, `encrypted` (PBS only), `notes` (first line only), `verification` (PBS only; object `{state, upid}` with `state` serialized lowercase as `"ok"` or `"failed"`, per `pbs-api-types/src/datastore.rs` `VerifyState` / `SnapshotVerifyState`).

**Aggregate agent-side, do not ship the raw listing.** A PBS datastore with thousands of snapshots would otherwise put thousands of entries on every tick. Emit one entry per `(vmid, storage)`: the latest `ctime`, its `volid`, its `size`, its `verification` (when present), its `protected`/`encrypted`, and a `count`. Bounded by guests x storages-with-backups, and the server only needs the latest.

**List shared storages once.** A PBS/NFS/CIFS storage is the same content from every node; listing it per node multiplies the most expensive call by the node count. `_storage_config_map` already fetches the datacenter `/storage` config and its `shared` flag: list a shared storage from the first node where it is active and emit `node: null`, list local storages per node with `node` set.

### 2. vzdump task outcomes

`GET /nodes/{node}/tasks?typefilter=vzdump&limit=<cap>` (default `source` = finished tasks). Fields: `upid`, `node`, `id`, `starttime`, `endtime`, `status`, `user`. Pass `status` **verbatim, capped**: it is a free string (`OK`, `WARNINGS: <n>`, or an error message), and the server classifies it, not the agent.

Two facts to bake into the contract, from `pve-manager` `PVE/API2/VZDump.pm:175-178`:
```perl
my $taskid;
$taskid = $local_vmids->[0] if scalar(@{$local_vmids}) == 1;
return $rpcenv->fork_worker('vzdump', $taskid, $user, $worker);
```
So `id` is the vmid **only when exactly one guest was backed up**; a scheduled multi-guest job has an empty `id`, and the job id is not in the task list at all (per-guest results live in the task log text, which we deliberately do not fetch). Tasks are therefore a **job-level** signal: "last night's backup on pve2 failed with X", never "VM 101 failed". The per-guest verdict comes from read 1.

### 3. Backup job definitions

`GET /cluster/backup` (one cluster-wide call). Keep: `id`, `enabled`, `schedule`, `storage`, `vmid`, `all`, `exclude`, `pool`, `node`, `mode`, `comment`, `next-run`. Drop the rest (bwlimit, compress, mailto, hooks...). This is display context ("VM 101 is backed up by job X to storage Y on schedule Z") and the schedule the server can later use to size a sensible default threshold.

`/cluster/backup/{id}/included_volumes` is deliberately **not** read: it is an ExtJS tree, and read 4 answers the question it would serve.

### 4. Guests covered by no job

`GET /cluster/backup-info/not-backed-up` (one cluster-wide call): `[{vmid, name, type}]`. This is the signal a job-level webhook can never produce: a guest that quietly dropped out of every job.

## Payload contract

A new key under `data["proxmox"]`, next to `storage`:

```json
"backups": {
  "age_s": 37,
  "guest_backups": [
    {"vmid": 100, "storage": "pbs-main", "node": null,
     "latest_ctime": 1789950000, "latest_volid": "pbs-main:backup/vm/100/2026-09-20T02:00:00Z",
     "latest_size": 8589934592, "count": 14,
     "verification": {"state": "ok", "upid": "UPID:..."}, "protected": false, "encrypted": true},
    {"vmid": 101, "storage": "local", "node": "pve2",
     "latest_ctime": 1789863600, "latest_volid": "local:backup/vzdump-lxc-101-2026_09_19-02_00_00.tar.zst",
     "latest_size": 1073741824, "count": 3}
  ],
  "tasks": [
    {"upid": "UPID:pve1:...:vzdump::root@pam:", "node": "pve1", "id": null,
     "starttime": 1789948800, "endtime": 1789950100, "status": "OK", "user": "root@pam"},
    {"upid": "UPID:pve2:...:vzdump::root@pam:", "node": "pve2", "id": null,
     "starttime": 1789948800, "endtime": 1789948900, "status": "job errors", "user": "root@pam"}
  ],
  "jobs": [
    {"id": "backup-6b2f1a7c-e1d3", "enabled": true, "schedule": "02:00", "storage": "pbs-main",
     "vmid": "100,101", "all": false, "exclude": null, "pool": null, "node": null,
     "mode": "snapshot", "comment": null, "next_run": 1790035200}
  ],
  "not_backed_up": [{"vmid": 105, "name": "scratch", "type": "qemu"}],
  "errors": []
}
```

Rules, in the order they matter:

1. **Three-level null, the OpenVPN posture.** Key **absent** = an agent older than this change (the server reads "unknown", never "no backups"). `backups: null` = the block could not be built at all (node listing failed). An **object with `errors`** = a partial read: every failed sub-read is an entry `{"scope": "storage"|"tasks"|"jobs"|"not_backed_up", "node": ..., "storage": ..., "message": ...}` (message capped at `_ERROR_MAX_LEN`). The server's rule for a partial read is asymmetric: a guest with a fresh backup on a successfully listed storage is fine (evidence exists), but a guest with **no** backup found is `unknown`, never `never`/`aged`, because its backups may live on the storage that failed. That rule only works if the agent reports **which** storage failed, so per-storage errors are load-bearing, not diagnostics.
2. **Not a sixth `collection` flag.** On the server `proxmox_cluster_reporters.completeness_score` is a STORED GENERATED column over the five flags that elects the single authoritative reporter; a sixth bit means a migration, re-weighted election, and a rollout window where every not-yet-upgraded reporter loses authority to whichever node upgraded first. Follow the `_storage_config_map` precedent (its docstring: "deliberately NOT threaded through the `collection` flags block, so it never flips storage_ok"): this block never touches `collection`, and carries its own failure signal per rule 1.
3. **Not every tick.** The content listing on a large PBS datastore is the most expensive call this collector will make, and backup age has a granularity of hours. Compute the block through the shared `TTLCache` (`fivenines_agent/cache.py`) with a fixed TTL (proposal: 600 s) and `store_if` rejecting `None`, so a recovered API is reported promptly instead of serving a stale failure. Emit `age_s` = seconds since the cached block was computed (monotonic), the `subscribed_age_s` / WireGuard-handshake-age pattern: the server anchors freshness as `received_at - age_s` and never trusts the agent's wall clock. `latest_ctime` and task times are data timestamps from PVE and pass through as-is.
4. **No new config key.** The TTL and the caps are constants. `config["proxmox"]` is unpacked with `pass_kwargs` into `proxmox_metrics(**config)`, so a key the server starts sending is a `TypeError` on every older agent and kills the whole Proxmox collector there (the config-splat hazard already documented on the server side).
5. **Caps and scrubbing.** Hard caps on every array (proposal: 5000 `guest_backups`, 50 tasks per node, 200 jobs, 2000 `not_backed_up`); when a cap trims, say so in `errors` with `scope: "cap"` rather than silently truncating, because a trimmed `guest_backups` reads as "never backed up" for whoever fell off the end. Every string field is customer-controlled (`notes`, `comment`, `status`, `volid`): `.encode("UTF-8", invalid="replace", undef="replace")` + `.delete("\u0000")` before it goes on the wire, the scrub every server ingester applies and Postgres refuses to store without.
6. **Skip entries without `ctime` or `vmid`** (both optional in the schema); they cannot feed a latest-per-guest computation and must not produce a `latest_ctime: null` entry the server would have to special-case.

## Fixture and tests

- `tests/fixtures/proxmox_contract_payload.json` is the shared contract with the server (asserted end-to-end by `test_contract_fixture_scenarios` with only `proxmoxer.ProxmoxAPI` mocked; the server's `spec/fixtures/proxmox_contract_payload.json` is a byte-identical copy). Every existing scenario's `payload` gains the `backups` key, and the server copy moves in lockstep in the companion PR. Add scenarios: `backups_ok` (two nodes, one shared PBS listed once + one local dir per node, one job, one not-backed-up guest, one `WARNINGS: 1` task), `backups_storage_partial` (one storage listing raises -> `errors` entry, other guests still present), `backups_block_failed` (`backups: null`), `backups_capped`.
- `make_proxmox_mock` in `tests/test_proxmox.py` gains the corresponding kwargs (`content_by_node_storage`, `content_raises_for`, `tasks_by_node`, `tasks_raises_for`, `backup_jobs`, `backup_jobs_raises`, `not_backed_up`, `not_backed_up_raises`).
- One test per rule above, and a `collect()` test proving a `backups` failure leaves `collection` untouched (rule 2).
- A TTL test proving the second `collect()` inside the TTL performs zero content-listing calls and that `age_s` grows.

## Out of scope

- Fetching task logs (`/nodes/{node}/tasks/{upid}/log`) for per-guest attribution: text parsing of a job-level log, fragile, and read 1 answers the per-guest question from evidence rather than from a report.
- Anything PBS-native (datastore usage, verify/sync/prune/GC job history). The `verification` pass-through in read 1 is the one PBS fact PVE hands us for free.
- A per-backup series. Latest-per-guest is what the alert needs; history can come later without changing this block.

## Acceptance criteria

- [ ] With a token created exactly as the setup guide says (`PVEAuditor` at `/`, `--privsep 0`), all four reads return data on a real PVE: one curl each, output pasted in the PR.
- [ ] `data["proxmox"]["backups"]` follows the contract above; the fixture scenarios exist on both sides and the server copy is byte-identical.
- [ ] A shared storage appears once in `guest_backups` (with `node: null`), a local storage once per node.
- [ ] A failing storage listing produces an `errors` entry naming the storage and does not remove guests found on other storages; a failing node listing produces `backups: null`; `collection` is untouched in both cases.
- [ ] A payload with thousands of snapshots on one datastore stays bounded (one entry per guest x storage) and the listing is not repeated inside the TTL.
- [ ] `status` reaches the wire verbatim (`OK`, `WARNINGS: 1`, an error string), capped, NUL-scrubbed; `id` is `null` for a multi-guest job task.
- [ ] No new key under `config["proxmox"]`.

spuyet:
## Summary

Adds `data["proxmox"]["backups"]`, the agent half of phase 2 of Five-Nines-io/fivenines_server#1159, so the dashboard can answer *"has every VM/CT had a backup in the last N hours, whatever the target?"* and *"which guests are in no backup job?"*. Closes #156.

**What it collects**, all off the existing `nodes -> storages` loop and all under the read-only `PVEAuditor` role the setup guide already provisions:
- **Per-guest latest backup per storage**, aggregated agent-side to one entry per `(vmid, storage)`: `{vmid, storage, node, latest_ctime, latest_volid, count, protected}`. A shared storage is listed once (`node: null`) and retried on the next active node after a failure; a local one per node.
- **Finished vzdump tasks per node** (job-level; `status` verbatim + capped; `id` null for a multi-guest job).
- **Backup job definitions** (`/cluster/backup`) and **guests in no job** (`/cluster/backup-info/not-backed-up`).

**Contract and safety**
- Three-level null (key absent / `null` / object with per-scope `errors[]`); never a sixth `collection` completeness flag; **no new key under `config["proxmox"]`**.
- Computed at most once per 600s through the shared `TTLCache` (keyed on host, port, token_id, verify_ssl) and re-emitted with `age_s`; one build bounded by a 30s wall-clock deadline.
- Every array capped (including `errors[]`) with a `scope: "cap"` error on trim; a job's `vmid`/`exclude` list is comma-truncated so a cap can't invent a phantom guest; every customer string scrubbed (slice-bounded, redacted on the log path); a non-list envelope is a scoped error or `backups: null`, never a silent empty read.

## Reading backups under a read-only token (the key design point)

Backup volumes are read from PVE's **prune-preview** endpoint, a dry run that changes nothing:

```
GET /nodes/{node}/storage/{storage}/prunebackups?prune-backups=keep-all=1
```

The first pass of this PR used the storage **content** listing (`?content=backup`), and testing on a real cluster showed that endpoint filters backup volumes per-volume: a `PVEAuditor` (audit-only) token gets HTTP 200 with the backups **removed**, so it would have needed a write-class `Datastore.Allocate` grant on every customer's token. An outside-model review (Codex) of that conclusion found the better path: the prune preview returns the **same present-volume evidence** (`volid`, `ctime`, `vmid`, `mark`) under plain `Datastore.Audit`, with no per-volume filter and no write privilege. Both facts were confirmed on the same live cluster (guest 114 on `node2/local`):

```
# PVEAuditor token, content listing -> filtered empty
GET .../storage/local/content?content=backup            -> 200 {"data":[]}

# root, content listing -> sees it (proves the volume exists)
pvesh get /nodes/node2/storage/local/content --content backup
  [{"vmid":114,"volid":"local:backup/vzdump-qemu-114-...vma.zst","ctime":1790004475,...}]

# PVEAuditor token, prune preview -> sees it, no privilege change
GET .../storage/local/prunebackups?prune-backups=keep-all=1
  {"data":[{"vmid":114,"mark":"keep","type":"qemu","ctime":1790004475,"volid":"local:backup/vzdump-qemu-114-...vma.zst"}]}   -> 200
```

So the whole feature works under `PVEAuditor` with **no token change for any existing customer**, and there is no rollout/security decision to make. The prune preview carries fewer fields than the content listing (no `size`, `verification`, `encrypted`); `protected` comes from the prune `mark`, and the missing fields are a later PBS-native or allocate-gated enrichment. A storage whose backups genuinely can't be read (for example a PBS datastore whose credentials lack prune rights) is a `storage`-scoped error, so those guests read `unknown`, never a false "not backed up".

The three other reads (`/cluster/backup`, `/cluster/backup-info/not-backed-up`, `/nodes/{n}/tasks?typefilter=vzdump`) also work under `PVEAuditor`, verified on the same cluster.

## Test Coverage

`fivenines_agent/proxmox.py`: **100% statement coverage**. Full suite: **3015 passed, 5 skipped**. mypy and flake8 clean; ASCII-only.

Coverage spans the three-level null contract, the TTL cache (zero reads inside the window, `age_s` growth, verify_ssl invalidation), the 30s deadline, shared-storage dedup + retry, every cap, malformed-envelope guards, CSV-safe vmid truncation, the log-path redaction, and the prune-preview read (protected from `mark`, undated-volume incompleteness).

## Pre-Landing Review

Five specialist passes (testing, maintainability, security, performance, simplification) plus Claude and Codex adversarial and structured reviews, all on Opus. No critical findings survived. Fixed along the way:
- **Security**: log-path text redacted + control-char-stripped (`_log_safe`), with a pre-redact cap so an unbounded PVE error body can't be a CPU sink on the watchdog loop; `errors[]` fields bounded.
- **Correctness (Codex)**: a job `vmid` char-cap that could split an id into a phantom guest (comma-safe truncation); malformed non-list envelopes masquerading as empty (now null / scoped error); a shared-storage failure suppressing a healthy node (now retries); `errors[]` unbounded (now capped); `verify_ssl` in the cache key; the task `limit` making its cap unreachable (now requests limit+1).
- **Design (Codex)**: the `/content` vs `/prunebackups` finding above, which removed the entire `Datastore.Allocate` requirement.

## Plan Completion

All of issue #156's reads, payload rules, and fixture/test requirements are implemented. The one acceptance criterion that could not hold as written -- "all four reads return data under `PVEAuditor`" -- is now **met**: switching the volume read to the prune-preview endpoint makes all four work under `PVEAuditor`.

## Verification Results

All four reads exercised on a live PVE with a `PVEAuditor` token: `/cluster/backup`, `/cluster/backup-info/not-backed-up`, and `/nodes/{n}/tasks?typefilter=vzdump` returned data; `/prunebackups` returned the backup volume with its `ctime` (see above). No token change required.

## TODOS

`TODOS.md` now carries a one-line pointer: the agent half is done here, everything server-side is in fivenines_server#1164, and the `Datastore.Allocate` question is reduced to the OPTIONAL later enrichment (verification/size) the prune preview doesn't carry.

## Documentation

README.md and CLAUDE.md updated: the Proxmox **Backups** section describes the block, and both now state the feature works under `PVEAuditor` via the prune-preview read (no token change), correcting the earlier draft that required `Datastore.Allocate`. No `CHANGELOG.md`/`VERSION` file in this repo; version is in `pyproject.toml` (1.19.0).

## Test plan
- [x] `poetry run pytest`: 3015 passed, 5 skipped.
- [x] `poetry run pytest tests/test_proxmox.py --cov=fivenines_agent.proxmox`: 100% coverage.
- [x] `flake8 --ignore=W503,E501` and `mypy` on the changed files: clean.
- [x] All four PVE reads exercised on a live cluster under a `PVEAuditor` token.


spuyet:
Server companion for phase 2 filed as Five-Nines-io/fivenines_server#1164, written against this PR's real wire contract (the `backups_contract` fixture block + the `backups_*` scenarios). It carries forward your permission finding as an explicit founder decision: Tier A (not-in-any-job + job-level task failures) ships under an unchanged `PVEAuditor`; Tier B (per-guest backup age) needs the `Datastore.Allocate` grant + a `SetupGuides::PROXMOX` change.

